PR #1, opened 2014-12-05
A mathematical morphology framework for Imglib2.

tinevez:
This PR ships a small mathematical morphology framework for Imglib2.

Rather than documenting what it does and how to use it, I put documentation and benchmarks on [Imglib2 wiki](http://fiji.sc/Imglib2_Morphological_Operations).

Though mathematical morphology extends far beyond what is contained in this PR, I wish to subject it to the team scrutiny early. 

For instance, this PR contains several calls to deprecated classes, I did not know how to workaround.


ctrueden:
I am torn. This PR is still awesome, and extremely useful. Obviously, we want this functionality in `imglib2-algorithm`. But it is a monster to review because it does not follow best practice of each commit representing a working state (compiling with passing tests). Instead, it introduces compile and test errors and then fixes them later. This is extremely common but makes the history harder to understand and patches harder to review.

Historically, @dscho and I have pushed for a clean Git history, but the fact is that the amount of work it requires to do that after the fact can sometimes be daunting and difficult to justify. How many more hours or days should we demand be invested into this patch series just to end up with the same end result?

Maybe a good compromise here would be to squash the whole thing together, then split it up again by new class. That would probably only take 1-2 hours of additional effort, would be easier to review and understand, and would avoid a lot of intermediate baggage. What do you think, @tinevez?

tinevez:
This PR will be closed. 
Following @ctrueden advice, I rewrite its history in an hopefully clean fashion. 
The content of this PR (with some typos fixed) can now be found in the PR #2.

Issue #5, opened 2015-07-30
Get span of RectangleShape


Squareys:
Hi @tpietzsch !

I am currently working on some neighborhood operations code, especially for imagej-ops. I need to access the span of an existing `RectangleShape`, but neither [Shape](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/neighborhood/Shape.java) nor [RectangleShape](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/neighborhood/RectangleShape.java) have a simple way to access it.

I guess you can probably get the span over the bounding box of the neighborhood, but that does not seem very nice.

**Ideas:**
Common for all shapes is a bounding box. What would you thing of a method to retrieve this bounding box/interval for example?
Another idea could be to have every Shape subclass provide getters for their parameters. Not all shapes have a span, obviously.

If you tell me, how you would want a solution to this to look, I could implement it.

Greeting,
Squareys (Jonathan)

tpietzsch:
@Squareys I think both would be good: for every Shape to be able to provide a bounding box interval and adding subclass-specific methods to e.g. RectangleShape.

Regarding bounding box interval: [Neighborhood](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/neighborhood/Neighborhood.java) already has a ```public Interval getStructuringElementBoundingBox()``` method. Because Shape does not have fixed dimensionality it would have to be ```public Interval getStructuringElementBoundingBox(int numDimensions)``` for Shape.

It would be cool if you could work on this!

tpietzsch:
Let's leave issue this open until the generic bounding box interval is implemented

stelfrich:
@tpietzsch I have implemented the bounding box for `Shape`s, see #27.

From your last comment it is not entirely clear, if you were aiming for something even more general like a `BoundingBox` interface?

PR #11, opened 2015-09-07
Implement getters for construction parameters in all Shapes

Squareys:
Hello everybody, Hello @tpietzsch !

I implemented the changes suggested in #5.

Greetings, Squareys


dietzc:
I'm wondering if there could be a `public Interval getBoundingBox()` method for each `Shape`, instead of providing `getSpan` or `getRadius`. It just feels a bit more consistent?

Squareys:
@dietzc See #5 for previous discussion. Both could be provided. I only implemented the one, since it is directly relevant for what I'm working on in imagej-ops.

dietzc:
thanks @Squareys

tpietzsch:
thanks @Squareys

PR #13, opened 2015-10-19
Fix the SubpixelLocalization bias bug.

### LINKED ISSUE #12: SubpixelLocalization class biased positions for 3D images.
This bug is responsible for the TrackMate bug fiji/TrackMate#68.

The [SubpixelLocalization](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/localextrema/SubpixelLocalization.java) class does not work 'fairly' in 3D. The Y coordinates is off by a considerable factor when localizaing peaks, whereas the X dimension is not affected.

Here are some results on synthetic images. Below I vary the peak position in X from 0 to 1 with 0.1 increments. the Y position of the peak is fixed to be 0.4 + integer.

subpixel_loc_error_dy0 4_dz0

The problem is not seen in 2D.

tinevez:
The SubpixelLocalization class is based on quadratic fitting of the peak
it tries to localize. To retrieve an estimate of the position, we
compute the gradient and the hessian at the peak integer location.

To compute the gradient and hessian, we use a randomAccess on the peak
image, and iterate through dimensions.

It turns out that the randomAccess position was not restored at the peak
center after dealing with a dimension. Therefore, the next dimensions
were calculated off-peak.

Fixes #12.

Signed-off-by: Jean-Yves Tinevez <jean-yves.tinevez@pasteur.fr>

### DIFF
--- a/src/main/java/net/imglib2/algorithm/localextrema/SubpixelLocalization.java
+++ b/src/main/java/net/imglib2/algorithm/localextrema/SubpixelLocalization.java
@@ -442,7 +442,6 @@ public class SubpixelLocalization< P extends Localizable, T extends RealType< T 
 		final int n = p.numDimensions();
 
 		access.setPosition( p );
-
 		final double a1 = access.get().getRealDouble();
 		for ( int d = 0; d < n; ++d )
 		{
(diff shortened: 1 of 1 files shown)


tpietzsch:
great! Thanks JY!

PR #14, opened 2015-10-22
A JUnit test for the accuracy of SubpixelLocalization.

tinevez:
This test generates synthetic, noiseless images in 2D and 3D, containing
11 peaks which position are offset by  0.1 pixels in one dimension. The
test passes with an accuracy tolerance of 0.03 pixels if the localizer
works properly.

It aims is to detect bias introduced by errors like in #12.


tinevez:
Thanks @tpietzsch

tpietzsch:
thank you @tinevez! Awesome job on #12. This will enter history as the most thoroughly fixed bug ever... :thumbsup:

PR #15, opened 2015-10-22
A port of ImgLib1 Scalespace by S. Saalfeld and S. Preibisch to ImgLib2.

tinevez:
This is liberal port of the scale-space algorithm written for Imglib1 by Stephan Saalfeld and Stephan Preibisch to Imglib2. This PR is made for evaluation by others. 

I added the logic to suppress edge responses, following the SIFT strategy. 
However, I translated some classes that depends on deprecated ImgLib2 classes, notably the SimpleMultithreading logic. 

An interactive test for this algorithm can be found here: imglib/imglib2-tests#5.


tinevez:
@axtimwalde @StephanPreibisch could you have an eye on this?

axtimwalde:
@tinevez,  I do not have strong opinions on whether to merge this.  So I would say yes if @StephanPreibisch and @tpietzsch agree.  More comments:  I didn't like this code ever.  It names itself ScaleSpace although it is a DoG detector with filters described in the SIFT paper an related publications.  In the modern world, this should be implemented as an Op of Ops that are re-usable elsewhere.  It also doesn't implement the smart scaling in octaves as in SIFT which would be quite interesting to speed things up and make them less memory demanding.  So, if I understand correctly, the benefits over

https://github.com/imglib/imglib2-algorithm/tree/master/src/main/java/net/imglib2/algorithm/dog

are that it includes an edge response filter and a local non-maxima suppression but otherwise repeats code?  May be I am getting it all wrong...

dietzc:
even if I'm not the guy deciding anything here, I suggest that no matter what will be decided the two implementations of https://github.com/imglib/imglib2-algorithm/tree/master/src/main/java/net/imglib2/algorithm/dog and the one of @tinevez are somehow unified and share the same coding concepts (`static` factory classes, `ExecutionContext` can be provided from outside) etc.

tinevez:
@axtimwalde Damn you are right! This totally escaped me! I had this port boiling for so long...

tinevez:
Ok I played and read the DogDetection class, and indeed this PR just brings duplicate code for something that already exists. 
I now wish to take it away.

The only non-duplicated piece as stated by @axtimwalde is the edge response suppressor, but this can go elsewhere. In the Subpixel stuff maybe.

PR #16, opened 2016-01-20
Added integral images of squares

stelfrich:
This adds the option for integral images of higher order while keeping the API consistent. Not providing an order will generate the standard integral image by summing up the values.

### DIFF
--- a/src/main/java/net/imglib2/algorithm/integral/IntegralImg.java
+++ b/src/main/java/net/imglib2/algorithm/integral/IntegralImg.java
@@ -76,11 +76,19 @@ public class IntegralImg< R extends NumericType< R >, T extends NumericType< T >
 
 	protected final Converter< R, T > converter;
 
+	protected final int order;
+
 	public IntegralImg( final RandomAccessibleInterval< R > img, final T type, final Converter< R, T > converter )
+	{
+		this(img, type, converter, 1);
+	}
+
+	public IntegralImg( final RandomAccessibleInterval< R > img, final T type, final Converter< R, T > converter, final int order )
 	{
 		this.img = img;
 		this.type = type;
 		this.converter = converter;
+		this.order = order;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


StephanPreibisch:
Hi, thanks for the effort. Just a question, are you referring to what's explained here? https://en.wikipedia.org/wiki/Summed_area_table#Extensions

stelfrich:
> Just a question, are you referring to what's explained here? https://en.wikipedia.org/wiki/Summed_area_table#Extensions

I had stumbled across [this publication](http://staffhome.ecm.uwa.edu.au/~00082689/papers/Shafait-efficient-binarization-SPIE08.pdf). But yes, that Wiki page describes the same principles.

axtimwalde:
Nice! @StephanPreibisch @stelfrich are you aware of the 2011 pre-ImgLib2 [2D implementations in Fiji](http://fiji.sc/Integral_Image_Filters)? Is anybody working on the blockmatching part, I'd be interested in using this.

StephanPreibisch:
I am aware of it, but not working on it @axtimwalde. Would be great to port it though :)

StephanPreibisch:
@stelfrich great, did you run a test on it to make sure it still works right after the added changes?

stelfrich:
> @StephanPreibisch @stelfrich are you aware of the 2011 pre-ImgLib2 2D implementations in Fiji? 

I was not aware of it, but I will take a look. Concerning the block matching: I don't think anyone is working on it at the moment (but I might be mistaken). But it might come in handy.

> @stelfrich great, did you run a test on it to make sure it still works right after the added changes?

@StephanPreibisch I ran a test that is not included in this PR. Would you like me to add one that compares a computed integral image with expected outcome?

StephanPreibisch:
If you could add some sort of JUnit test would be great of course ... e.g. compute the sum of pixels at some random locations in the image via the Integral Image and "by hand"? Would that be too much to ask? I know, we also did not do it before, but it would be good to start :)

stelfrich:
@StephanPreibisch Definitely not too much to ask. I have implemented your idea, however, only for 2D for now.

StephanPreibisch:
Great, thanks so much! And I assume it also works ... :)

PR #17, opened 2016-01-27
Specialized cursor for RectangleNeighborhoods

stelfrich:
The `IntegralCursor` returns specific corner values of `RectangleNeighborhood`s that can be used for computations on integral images. 

In the 2D example the values at the following positions are returned by `IntegralCursor`:
* (neighMin, neighMin)
* (neighMax-1, neighMin)
* (neighMin, neighMax-1)
* (neighMax-1, neighMax-1)

I would rather have it return the values at the corner positions (and call it CornerCursor) but that renders it effectively useless for use with local operations on integral images. The problem is that for computing the mean in a `RectangleNeighborhood` of an image, values from a `RectangleNeighborhood` with `span+1` at the positions returned by the `IntegralCursor` are required.

### DIFF
--- a/src/main/java/net/imglib2/algorithm/integral/IntegralCursor.java
+++ b/src/main/java/net/imglib2/algorithm/integral/IntegralCursor.java
@@ -34,8 +34,6 @@
 
 package net.imglib2.algorithm.integral;
 
-import java.util.Vector;
-
 import net.imglib2.AbstractEuclideanSpace;
 import net.imglib2.Cursor;
 import net.imglib2.RandomAccess;
@@ -45,19 +43,23 @@ import net.imglib2.algorithm.neighborhood.RectangleNeighborhood;
  * A cursor implementation that returns specific corner values of
  * {@link RectangleNeighborhood}s.
  *
- * The cursor returns, for example in 2D, the values at the following
- * positions:
+ * The cursor returns, for example in 2D, the values at the following positions:
  *
- * <ol>
+ * <ul>
  * <li>(neighMin, neighMin),</li>
  * <li>(neighMax-1, neighMin),</li>
  * <li>(neighMin, neighMax-1), and</li>
  * <li>(neighMax-1, neighMax-1).</li>
- * </ol>
+ * </ul>
  *
- * The mechanism naturally extends to nD. The current position can be
- * obtained from {@code getVector()} with 0s encoding neighMin and 1s
- * encoding neighMax-1.
+ * The mechanism naturally extends to nD. The current position can be obtained
+ * from {@code getVector()} with 0s encoding (neighMin) and 1s encoding
+ * (neighMax-1). The iteration order follows the (binary-reflected) Gray code
+ * pattern such that only one dimension of the target position is modified per
+ * move.
+ * 
+ * @see <a href="http://en.wikipedia.org/wiki/Gray_code">http://en.wikipedia.org
+ *      /wiki/Gray_code</a>
  *
  * @author Stefan Helfrich
  */
@@ -65,6 +67,9 @@ public class IntegralCursor< T > extends AbstractEuclideanSpace implements Curso
 {
 
 	private int index = 0;
+
+	/** Reflected binary code (Gray code) **/
+	private int code = 0;
 
 	private final int maxIndex;
 
@@ -77,7 +82,7 @@ public class IntegralCursor< T > extends AbstractEuclideanSpace implements Curso
 		super( neighborhood.numDimensions() );
 		this.neighborhood = neighborhood;
 		source = neighborhood.getSourceRandomAccess();
-		maxIndex = ( int ) Math.round( Math.pow( 2, neighborhood.numDimensions() ) );
+		maxIndex = ( ( int ) Math.round( Math.pow( 2, neighborhood.numDimensions() ) ) ) - 1;
 		reset();
 	}
 
@@ -93,37 +98,49 @@ public class IntegralCursor< T > extends AbstractEuclideanSpace implements Curso
 	@Override
 	public void reset()
 	{
+		index = -1;
+
 		long[] min = new long[ neighborhood.numDimensions() ];
 		neighborhood.min( min );
 		source.setPosition( min );
-		source.bck( 0 );
-		index = 0;
 	}
 
 	@Override
 	public void fwd()
 	{
-		// Extract each dimension individually from currentPosition
-		final long[] cornerPosition = new long[ neighborhood.numDimensions() ];
-		for ( int d = 0; d < neighborhood.numDimensions(); ++d )
+		// Check if cursor is uninitialized
+		if ( index == -1 )
 		{
-			int valueInDimension = index >> d;
-			valueInDimension = valueInDimension & 1;
-
-			if ( valueInDimension == 1 )
-			{
-				// if bit in dimension is set
-				cornerPosition[ d ] = neighborhood.max( d ) - 1;
-			}
-			else
-			{
-				// if not
-				cornerPosition[ d ] = neighborhood.min( d );
-			}
+			index++;
+			return;
 		}
 
-		source.setPosition( cornerPosition );
-		++index;
+		/*
+		 * Adapted from Wikipedia:
+		 * 
+		 * To construct the binary-reflected Gray code iteratively, at step 0
+		 * start with the code = 0, and at each step index > 0 find the bit
+		 * position of the least significant 1 in the binary representation of
+		 * index and flip the bit at that position in the previous code to get
+		 * the next code.
+		 */
+		index++;
+
+		// Update Gray code
+		int mask = Integer.lowestOneBit( index );
+		code ^= mask;
+
+		// Move the cursor in the dimension of the updated bit
+		int updatedDimension = Integer.numberOfTrailingZeros( index );
+		int bitInDimension = ( code & mask ) >> updatedDimension;
+		if ( bitInDimension == 1 )
+		{
+			source.setPosition( neighborhood.max( updatedDimension ) - 1, updatedDimension );
+		}
+		else
+		{
+			source.setPosition( neighborhood.min( updatedDimension ), updatedDimension );
+		}
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


dietzc:
I didn't have time do to a close review (I think @tpietzsch is the best person to do the review here), but can you apply the imglib2 style for all your classes? it can be found in https://github.com/imglib/imglib2/tree/master/doc. This would be great!

stelfrich:
> but can you apply the imglib2 style for all your classes?

@dietzc I have applied a _Source > Format_ and _Source > Clean up_ in the latest commit with the settings applied in Eclipse Mars. Is that the desired outcome? I am asking because there were some discrepancies with my configuration (project specific settings don't necessarily seem to overwrite global settings..).

dietzc:
did you first set the checkstyle of imglib2-core?

stelfrich:
> did you first set the checkstyle of imglib2-core?

Here is what I did:

1. Right click imglib2-algorithm > Properties > Java Code Style > Clean Up > Import imglib2-eclipse-clean-up-style.xml
2. Right click imglib2-algorithm > Properties > Java Code Style > Formatter > Import imglib2-eclipse-mars-clean-up-formatter-style.xml
3. Source > Format
4. Source > Clean up

I have repeated the procedure just now and there's no changes (neglecting the whitespaces in the header).

dietzc:
It seems I took a look at the wrong commit. thanks.

stelfrich:
Short recap of a discussion with @dietzc and @axtimwalde in Gitter:

- [ ] Pass `Neighborhood` to `IntegralCursor`
- [ ] Move `IntegralCursor` to https://github.com/imglib/imglib2-algorithm/tree/integralCursor/src/main/java/net/imglib2/algorithm/integral

Was there a decision to rename `IntegralCursor` to `CornerCursor` or is that obsolete with the move to _src/main/java/net/imglib2/algorithm/integral_?

dietzc:
> Was there a decision to rename IntegralCursor to CornerCursor or is that obsolete with the move to src/main/java/net/imglib2/algorithm/integral?

I suggest to keep the name `IntegralCursor`, as it now lives in the very same package.

stelfrich:
I have implemented the changes minus the optional rename. I had to extend the interface of `RectangleNeighborhood` to get the `sourceRandomAccess` because `IntegralCursor` is no longer an inner class.

tpietzsch:
@stelfrich @dietzc Hi, I just had a look at it for the first time, sorry for long delay... To be honest, I don't know what the point of this is. Why piggyback on `RectangleNeighborhood`, not just make a new neighborhood (respectively shape) whose cursors iterate the rectangle corners (similar JY's `DiamondTipsNeighborhood`; could possibly extend `RectangleNeighborhood`)? I need more context. How do you want it to be used?

stelfrich:
> To be honest, I don't know what the point of this is

The `IntegralCursor` will most be used in conjunction with integral images for computations of mean, standard deviation, variance (hence the move to `net.imglib2.algorithm.integral` in the latest commit). For those computations, only 2^dim values from an integral image are required, which are exactly the values returned by the `IntegralCursor`.

Having a new neighborhood with this implementation as `LocalCursor` is possible but seems like a lot of code duplication to me (see `DiamondNeighborhood*` and `DiamondTipNeighborhood*` classes). With the proposed implementation, we are adding one class and a getter in `RectangleNeighborhood` instead.

Does this make sense?

tpietzsch:
> @tpietzsch The IntegralCursor will most likely only be used in conjunction with integral images for computations of mean, standard deviation, variance (hence the move to net.imglib2.algorithm.integral in the latest commit). For those computations, only 2^dim values from an integral image are required, which are exactly the values returned by the IntegralCursor.

Hmm, that much I guessed...  My question is more along the lines: where would it be used in code? Is this user facing or just used internally, e.g., by some (yet to be written) accessor that provides the mean, standard deviation, variance, for a specific window size?

> Having a new neighborhood with this implementation as LocalCursor is possible but seems like a lot of code duplication to me (see DiamondNeighborhood* and DiamondTipNeighborhood* classes). With the proposed implementation, we are adding one class and a getter in RectangleNeighborhood instead.

That depends on the answer to the above question. If it is intended to be exposed to users, I would go the extra-mile and create a proper accessors, accessible, etc. It could extend the RectangleShape stuff without all that much code duplication, although it would be certainly more that the single "IntegralCursor".

Independent of all this, regarding the actual code of `IntegralCursor`:

- Returning `Vector< Integer >` is bad. `ArrayList<Integer>` is better, and probably best is simply `int[]`.
- Because you increment `index` at the end of `fwd()`, `getCornerInteger` and `getCornerVector` will return wrong results.
- `fwd()` can most likely made much more efficient if you do not do a full `source.setPosition(long[])` every time but instead setPositions for the dimensions that change.
- Currently, everything that you do in `reset()` except `index = 0;` is pointless because of the full `setPosition` in `fwd()`. If `fwd()` is changed to be more efficient, this too has to be revised. Otherwise should be removed.

stelfrich:
> Is this user facing or just used internally

It is used internally in some ops. The user just decides that he wants, for example, to apply a local threshold to an image providing the name of the method, a `Shape`, and an `OutOfBoundsFactory`. In the end, there will be two implementations: a "default" one and one based on integral images when the `shape instanceof RectangleShape`. All in all, the computation of the integral image and the application of `Integral*` ops is hidden from the user.

> Independent of all this, regarding the actual code of IntegralCursor

Thanks for the comments, @tpietzsch.

tpietzsch:
> It is used internally in some ops. The user just decides that he wants, for example, to apply a local threshold to an image providing the name of the method, a Shape, and an OutOfBoundsFactory. In the end, there will be two implementations: a "default" one and one based on integral images when the shape instanceof RectangleShape. All in all, the computation of the integral image and the application of Integral* ops is hidden from the user.

Ok, that makes sense.

Regarding my comment
> fwd() can most likely made much more efficient if you do not do a full source.setPosition(long[]) every time but instead setPositions for the dimensions that change.

If you want to dig into this, have a look at `net.imglib2.interpolation.randomaccess.NLinearInterpolator`. Using gray code patterns you can make it work with only changing one dimension per `fwd()`. This will alter the order in which corners are visited, but algos should be able to deal with that

stelfrich:
> Independent of all this, regarding the actual code of `IntegralCursor`

I have addressed your comments in the latest commit @tpietzsch.

> Ok, that makes sense.

You can take a look at the [`IntegralMean`](https://github.com/stelfrich/imagej-ops/blob/dbd4c5f058201c117f2eacc7496d88c037b8fc3a/src/main/java/net/imagej/ops/stats/IntegralMean.java#L77-L88) implementation if you like. The computation is independent of the iteration order of the cursor and the dimension.

stelfrich:
Any news on this @tpietzsch? We could move the `IntegralCursor` to Ops if you feel that imglib is not the right place for it. If there are any concerns regarding the implementation - I'd be happy to address them.

I have implemented integral versions of local thresholding ops that benefit from (and rely on) this PR.

tpietzsch:
@stelfrich Did you test this and could you please also commit the tests?
Here are some comments:
- javadoc on iteration order https://github.com/imglib/imglib2-algorithm/pull/17/files#diff-03b370b9393245a7920a44a5739022a1R49 is not correct anymore.
- https://github.com/imglib/imglib2-algorithm/pull/17/files#diff-03b370b9393245a7920a44a5739022a1R103 ff are more verbose than necessary. `source` is `Positionable`, so `neighborhood.min( source );` will do the job.
- You do not reset `code` in `reset()` or copy it in the copy constructor.

These final ones are clearly bugs, which makes me wonder how thoroughly you tested this.

As a general remark (which probably no one will read buried in here, but anyway):
Please, do not mis-use pull-requests as a means of getting cheap code reviews and debugging. PRs should be for stuff that is about ready to get merged, not for trying ideas and evolving code that is still under heavy development. This would be better discussed and developed in an issue first.

tpietzsch:
@stelfrich Recently, @axtimwalde started to look into integral images, I wonder how far along that is and how it would influence this PR. Can you comment on progress @axtimwalde?

axtimwalde:
@tpietzsch @stelfrich During my attempt to implement an image integrator I found a conceptual issue in the Ops framework that @ctrueden and @dietzc are currently fixing.  Without this fix the implementation becomes unnecessarily verbose and / or copy intense.  I suggest that @stelrich's implementation including everything that it does and depends on goes into Ops.

tpietzsch:
@axtimwalde If your integral images will live in Ops, then that might make most sense indeed. Although I would also kind of like to have integral images in imglib...

acardona:
@tpietzsch There is an ImgLib2-based integral image in the scripting package, purely ImgLib2, written years ago ... and I am sure that, in any case, a new one could be rewritten in little time.

tpietzsch:
@acardona I know, that is already in imglib2-algorithm actually.
As far as I know @axtimwalde is trying to go beyond that, working toward "integral block images" that are split in blocks that each are an integral image. This would avoid that values in the integral image increase indefinitely (that would be especially relevant for one of our applications where pixel type is "Multi-set of Labels").

axtimwalde:
@cardona thanks, and yes we are aware of this code, for the record https://github.com/imglib/imglib2-script/tree/master/src/main/java/net/imglib2/script/algorithm/integral the idea here, though, was to port the entirety of the code from integral generation to consuming code into the ops framework.  Personally, I want an entirely different access concept and the opportunity to store integral cell images that is not covered here.  But I am starting to get really nervous about this green merge button here.  Can we close this PR?  It's just not the right place to discuss incomplete ideas.

acardona:
@tpietzsch Thanks for the explanations, integral block images sounds right. Been there, wrote the UnsignedLongType because of the issue.

tpietzsch:
@axtimwalde if we move the `IntegralCursor` to ops, there will still remain the addition of `getSourceRandomAccess()` to `RectangleNeighborhood` in this PR.
@stelfrich can you split that out into a new PR or modify this one?

stelfrich:
> can you split that out into a new PR or modify this one?

Will do.

> As a general remark (which probably no one will read buried in here, but anyway)

Put it in a _Contributing_ section of a README.md. And to be on the safe side, add a reminder in the first line of the default [PR template](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/).

tpietzsch:
@stelfrich 

> Put it in a Contributing section of a README.md. And to be on the safe side, add a reminder in the first line of the default PR template.

Thanks for the pointer! I didn't know about PR template

hinerm:
>Thanks for the pointer! I didn't know about PR template

Me neither. That's awesome! Thanks @stelfrich

ctrueden:
You can also [make a `CONTRIBUTING.md` file](https://github.com/blog/1184-contributing-guidelines) which is linked from the top of the PR creation screen. I agree with @stelfrich that this policy needs to be spelled out, since it is a social convention that varies by project. (Some projects, for example, encourage opening a PR as soon as possible so that all discussion can happen there.)

PR #18, opened 2016-02-22
Added CellProfiler 3x3 operations

LeeKamentsky:
@LeonYang5114 Hi Leon, this branch / pr has the basic morphological operations from CellProfiler. The most valuable parts are the tables, after that, there's enough code for me to see that it's working, but not much more than that. Feel free to modify as you see fit.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,9 +1,9 @@
-Copyright (c) 2009 - 2015, Tobias Pietzsch, Stephan Preibisch, Barry DeZonia,
-Stephan Saalfeld, Curtis Rueden, Albert Cardona, Christian Dietz, Jean-Yves
-Tinevez, Johannes Schindelin, Jonathan Hale, Lee Kamentsky, Larry Lindsey, Mark
-Hiner, Michael Zinsmaier, Martin Horn, Grant Harris, Aivar Grislis, John
-Bogovic, Steffen Jaensch, Stefan Helfrich, Jan Funke, Nick Perry, Mark Longair,
-Melissa Linkert and Dimiter Prodanov.
+Copyright (c) 2009 - 2016, Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+Aivar Grislis, Jonathan Hale, Grant Harris, Stefan Helfrich, Mark Hiner,
+Martin Horn, Steffen Jaensch, Lee Kamentsky, Larry Lindsey, Melissa Linkert,
+Mark Longair, Brian Northan, Nick Perry, Dimiter Prodanov, Curtis Rueden,
+Johannes Schindelin, Jean-Yves Tinevez and Michael Zinsmaier.
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification,
--- a/src/main/java/net/imglib2/algorithm/Algorithm.java
+++ b/src/main/java/net/imglib2/algorithm/Algorithm.java
@@ -2,12 +2,12 @@
  * #%L
  * ImgLib2: a general-purpose, multidimensional image processing library.
  * %%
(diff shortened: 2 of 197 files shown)


LeonYang5114:
Thanks for letting know. I will look into that.

ctrueden:
Awesome, thanks @LeeKamentsky!

I guess unit tests are left as an exercise to the reader (i.e., @LeonYang5114), eh? :grin: 

@tinevez @tpietzsch Any thoughts on this?

tinevez:
Hi all.
I cannot peek on this properly before a long time ... sorry.

tpietzsch:
Because it's only 2D, it should be moved to a sub-package `net.imglib2.algorithm.morphology.twod` or similar or reflected in the class names.
`Abstract3x3TableOperation` looks good. The other ones need at least a class javadoc. It's impractival to figure out what they do from looking at the boolean tables or googling and guessing.

LeeKamentsky:
Here's the description for `Thicken` from the Python code:

    Thicken the objects in an image where doing so does not connect them
    
    0 0 0    ? ? ?
    0 0 0 -> ? 1 ?
    0 0 1    ? ? ?
    
    1 0 0    ? ? ?
    0 0 0 -> ? 0 ?
    0 0 1    ? ? ?

A single round of `Thicken` is the same as a dilation with a 3x3 structuring element except for cases like this:

    1 1 0 1 1
    1 1 0 1 1   Don't fill in centers because it would merge unconnected components.
    1 1 0 1 1

    1 1 1 1 1    1 1 1 1 1    1 1 1 1 1
    1 1 0 1 1    1 1 1 1 1    1 1 1 1 1
    1 1 0 1 1 -> 1 1 0 1 1 -> 1 1 1 1 1 It takes two rounds to fill in holes like this,
    1 1 0 1 1    1 1 1 1 1    1 1 1 1 1 but it does happen eventually
    1 1 1 1 1    1 1 1 1 1    1 1 1 1 1

LeeKamentsky:
Regarding spur1 and 2 and thin1 and 2, it's almost as if they should be private or otherwise hidden - they are both meant to be used in tandem. I would write tests for Spur and Thin and let those tests provide coverage for the other functions.

And it looks like the tests for Thin will fail!!! Because it's not there!!! Let me see if I can fix that quickly and maybe I can write the tests for Spur and Thin too.

LeonYang5114:
I have added the remaining tests and added javadoc for each class using the information provided on http://www.broadinstitute.org/~leek/CPmanual/Morph.html 

The abstract class is modified a little to make it simpler. The classes are moved to a subpackage, as suggested by @tpietzsch I pick the package name of `table2d` because I think they are not only 2d, but also backed by lookup table.

Since they are now in a subpackage, the previous dependency on `MorphologyUtils` have to be removed (as it util class has package visibility). It seems that the `MorphologyUtils#getSuitableFactory`, which allows non-`NativeType` type and could returns `ListImgFactory`, is general enough to be put in `net.imglib2.util.Util`.

The history of this branch is messy right now. @LeeKamentsky feel free to do a rebase to clean it up, or if you give me the power, I could do it.

Any comment?

LeeKamentsky:
@LeonYang5114 just one comment, Thanks. The improvements to the code are great and the unit tests look good too. You should feel free to do the rebase if that makes things more comprehensible. If there's something I need to do to hand things off to you, let me know, otherwise I am more than happy to let you do whatever is necessary.

LeonYang5114:
@LeeKamentsky I clean up the history, so we only have 2 commits now: one for adding the implementations, the other for the tests.

@tpietzsch Could you check if anything need to be changed?

ctrueden:
I did a little cleanup. I think it LGTM :+1:
@tpietzsch You see any issues with it?

LeeKamentsky:
Thx Curtis.

tpietzsch:
thanks guys!
@ctrueden I would have prefered the header updates in a separate PR.

PR #19, opened 2016-03-14
RectangleNeighborhood: add getter for source RA

stelfrich:
See https://github.com/imglib/imglib2-algorithm/pull/17#issuecomment-196358317.


stelfrich:
Thanks @tpietzsch!

PR #20, opened 2016-03-22
Change visibility of MorphologyUtils methods

LeonYang5114:
Methods with package visibility are changed to be public, so that other packages/projects could use.

### DIFF
--- a/src/main/java/net/imglib2/algorithm/morphology/MorphologyUtils.java
+++ b/src/main/java/net/imglib2/algorithm/morphology/MorphologyUtils.java
@@ -52,7 +52,7 @@ public class MorphologyUtils
 	 *         source image.
 	 *         </ol>
 	 */
-	static final < T > long[][] computeTargetImageDimensionsAndOffset( final Interval source, final Shape strel )
+	public static final < T > long[][] computeTargetImageDimensionsAndOffset( final Interval source, final Shape strel )
 	{
 		/*
 		 * Compute target image size


ctrueden:
This helps the Ops project in particular, which otherwise has to recapitulate this code in its own utility class.

tpietzsch:
@tinevez are you ok with this?

tinevez:
The methods you make visible are particularly ugly, notably the subAA and so on. Are YOU ok with this?

tpietzsch:
I have no opinion, I'm not familiar with the code. imglib2-algorithms is not out of beta, so I have no big reservations about adding public API that may be removed later. @ctrueden @LeonYang5114 maybe you could elaborate on why this is required exactly?

LeonYang5114:
I am migrating the morphology code from imglib2-algorithm to imagej-ops, and actually `computeTargetImageDimensionsAndOffset` is needed in handling the padding and alignment. The method is not complicated, but it would be nice to reuse code.

I make others methods also public just because I thought that it might make things consistent. I will modify the PR so that only the method I need is public, if you want.

I recognize that some methods in that utility class is not specific to morphology e.g. `copyCropped`, `subXXX`, and some are almost duplicate, e.g. `getSuitableFactory` vs `net.imglib2.util.Util#getArrayOrCellImgFactory`, `getTypeFromInterval` vs `net.imglib2.util.Util#getTypeFromInterval`. This utility class might need some rework.

tpietzsch:
> I make others methods also public just because I thought that it might make things consistent. I will modify the PR so that only the method I need is public, if you want.

Yes, I would prefer that.

LeonYang5114:
Please check if this works.

PR #21, opened 2016-04-21
Flood Fill for RandomAccessible

hanslovsky:
Implemented Flood Fill for RandomAccessible:
Starting at seed point within randomAccessible, the current pixel gets filled with newLabel. Neighboring pixels, defined by shape, get added to the "queue" if they are part of the same connected component and have not been visited yet, according to fillPolicy.
Implemented FillPolicy for the most simple case of net.imglib2.type.Type in GenericTypeFillPolicy, including factory.
Includes unit test.

### DIFF
--- a/src/main/java/net/imglib2/algorithm/fill/FillPolicy.java
+++ b/src/main/java/net/imglib2/algorithm/fill/FillPolicy.java
@@ -1,0 +1,17 @@
+package net.imglib2.algorithm.fill;
+
+import net.imglib2.Localizable;
+import net.imglib2.RandomAccessible;
+import net.imglib2.algorithm.neighborhood.Shape;
+
+/**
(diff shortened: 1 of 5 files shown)


axtimwalde:
Cool.  Suggested changes as discussed:
* license blurb
* Comment must be "interative n-dimensional floodfill for arbitrary neighborhoods"
* add me as co-author
* add infrequent (e.g. i == 100000) cleanup of candidate list to prevent unnecessary memory buildup
* reconsider FillPolicy and replace by Comparator<Pair<A,B>> and Write<B> with default implementation for A, B extending Comparable and B extending Type
* invokation methods for generic case with immutable seed A and fill color B on RandomAccessible<A> srouce and RandomAccessible<B> target with both interfaces being passed parameters and convenience method for A and B implementing Comparable and B implementing Type

axtimwalde:
Then rebase and file as new pull request when ready for merge.

PR #22, opened 2016-04-23
Iterative n-dimensional floodfill for arbitrary neighborhoods

hanslovsky:
Flood fill based on user defined connectivity check and only if pixel
has not been visited yet (Comparator). User defines how to write into
target using Writer and neighborhood using Shape. That way, we do not
have any  requirement on source and target types. Added convenience
methods for input type extending Comparable (and Type) and output
extending Comparable and Type.

Added default implementations for Comparator and Writer.

Update of #21 according to [comments by @axtimwalde](https://github.com/imglib/imglib2-algorithm/pull/21#issuecomment-213164965).
Requires latest checkout of imglib2 master.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,12 +5,12 @@
 	<parent>
 		<groupId>net.imglib2</groupId>
 		<artifactId>pom-imglib2</artifactId>
-		<version>6.0.0</version>
+		<version>7.1.0</version>
 		<relativePath />
 	</parent>
 
 	<artifactId>imglib2-algorithm</artifactId>
-	<version>0.4.1-SNAPSHOT</version>
+	<version>0.5.0-SNAPSHOT</version>
 
 	<name>ImgLib2 Algorithms</name>
 	<description>Useful image processing algorithms.</description>
@@ -28,7 +28,6 @@
 		<dependency>
 			<groupId>net.imglib2</groupId>
 			<artifactId>imglib2</artifactId>
-			<version>2.8.1-SNAPSHOT</version>
 		</dependency>
 		<dependency>
 			<groupId>net.imglib2</groupId>
@@ -75,16 +74,6 @@
 						</manifest>
 					</archive>
 				</configuration>
-			</plugin>
-			<!-- Do not enforce fail on SNAPSHOT dependencies -->
-			<plugin>
-				<artifactId>maven-enforcer-plugin</artifactId>
-				<executions>
-					<execution>
-						<id>enforce-rules</id>
-						<phase>none</phase>
-					</execution>
-				</executions>
 			</plugin>
 		</plugins>
 	</build>
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,7 @@
 		<dependency>
 			<groupId>net.imglib2</groupId>
 			<artifactId>imglib2</artifactId>
+			<version>2.8.1-SNAPSHOT</version>
 		</dependency>
 		<dependency>
 			<groupId>net.imglib2</groupId>
@@ -48,6 +49,11 @@
 		<dependency>
 			<groupId>gov.nist.math</groupId>
 			<artifactId>jama</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>net.sf.trove4j</groupId>
+			<artifactId>trove4j</artifactId>
+			<version>3.0.3</version>
 		</dependency>
 
 		<!-- Test dependencies -->
@@ -70,6 +76,17 @@
 					</archive>
 				</configuration>
 			</plugin>
+			<!-- Do not enforce fail on SNAPSHOT dependencies -->
+			<plugin>
+				<artifactId>maven-enforcer-plugin</artifactId>
+				<executions>
+					<execution>
+						<id>enforce-rules</id>
+						<phase>none</phase>
+					</execution>
+				</executions>
+			</plugin>
 		</plugins>
 	</build>
+
 </project>
--- a/src/main/java/net/imglib2/algorithm/fill/ComparableComparator.java
+++ b/src/main/java/net/imglib2/algorithm/fill/ComparableComparator.java
@@ -1,0 +1,61 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (c) 2009 - 2016, Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+ * John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+ * Aivar Grislis, Jonathan Hale, Philipp Hanslovsky, Grant Harris, Stefan Helfrich,
+ * Mark Hiner, Martin Horn, Steffen Jaensch, Lee Kamentsky, Larry Lindsey,
+ * Melissa Linkert, Mark Longair, Brian Northan, Nick Perry, Dimiter Prodanov,
+ * Curtis Rueden, Johannes Schindelin, Jean-Yves Tinevez and Michael Zinsmaier.
+ * All rights reserved.       
+ *                            
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *                            
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *                            
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *                            
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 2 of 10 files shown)


pom.xml
```diff
@@ -28,6 +28,7 @@
 		<dependency>
 			<groupId>net.imglib2</groupId>
 			<artifactId>imglib2</artifactId>
+			<version>2.8.1-SNAPSHOT</version>
```
axtimwalde: We cannot merge that until there is a non-SNAPSHOT release of ImgLib2 or @ctrueden will kill you.  @tpietzsch @ctueden @StephanPreibisch are you ok with releasing imglib2-2.9.0 quickly?  This release should contain the `RandomAccessiblePair` and @tpietzsch's `Views` fix https://github.com/imglib/imglib2/pull/129.

pom.xml
```diff
@@ -70,6 +76,17 @@
 					</archive>
 				</configuration>
 			</plugin>
+			<!-- Do not enforce fail on SNAPSHOT dependencies -->
```
axtimwalde: That will have to go with depending on a non-SNAPSHOT release.  It's ok while hacking, not ok for a merge to master.

src/main/java/net/imglib2/algorithm/fill/ComparableComparator.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ */
+public class ComparableComparator< T extends Comparable< T >, U extends Comparable< U >  >
+        implements Comparator<Pair< T, U > > {
+
+    @Override
+    public int compare( final Pair< T, U > p1, final Pair< T, U > p2) {
+        int comp1 = p1.getA().compareTo( p2.getA() );
+        int comp2 = p1.getB().compareTo( p2.getB() );
+        return  comp1 == 0 ?
```
axtimwalde: Please replace by
`return ( comp1 | comp2 ) != 0;

hanslovsky: I like the idea of a bitwise operator here, but shouldn't it be xor?
```
| comp1 | comp2 | (comp1 OR comp2) != 0 | (comp1 XOR comp2) != 0 | Desired Result                          |
|-------+-------+-----------------------+------------------------+-----------------------------------------|
| 0     |     0 | False                 | False                  | False                                   |
| 0     |     1 | True                  | True                   | True                                    |
| 1     |     0 | True                  | True                   | True                                    |
| 1     |     1 | True                  | False                  | False (not in same connected component) |
```

hanslovsky: Actually, my thoughts on XOR are only true, if comp1 and comp2 are binary, i.e. they always take the same value if non-zero. That makes this a little more copmlicated.

src/main/java/net/imglib2/algorithm/fill/ComparableComparator.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @author Philipp Hanslovsky &lt;hanslovskyp@janelia.hhmi.org&gt;
+ *
+ *
+ * Default implementation of {@link Comparable} for the use in {@link FloodFill} when source and target are both
+ * {@link net.imglib2.RandomAccessible<Comparable>}
+ *
+ */
+public class ComparableComparator< T extends Comparable< T >, U extends Comparable< U >  >
```
axtimwalde: I think we also want an `EqualsComparator` that works with `Object.equals`.  That would be the most generic thing possible with a simple Interface.  The bad news is that neither of the ImgLib2 `Types` overrides `equals` in a sensible way, and so all the generalization effort leads nowhere.  `Comparable<T>` implies an order in the value domain of `T`, i.e. we cannot use it for `ARGBType`.  I am currently working on a branch of ImgLib2 that will implement `Type.valueEquals` as an alternative to `Object.equals` which is used in other place s with different assumptions.

src/main/java/net/imglib2/algorithm/fill/TypeWriter.java
```diff
@@ -0,0 +1,54 @@
...
+import net.imglib2.type.Type;
+
+/**
+ * @author Philipp Hanslovsky &lt;hanslovskyp@janelia.hhmi.org&gt;
+ *
+ *
+ * Default implementation of {@link Writer} for {@link Type}.
+ *
+ */
+public class TypeWriter< T extends Type< T > > implements Writer< T > {
```
axtimwalde: I love that classname!

hanslovsky: Haha, thank you. I only realized after implementaiton but that was a pleasent surprise.

src/main/java/net/imglib2/algorithm/fill/ComparableComparator.java
axtimwalde: Depends on the logic that you want to express.  The name of the class suggests that you implement comparison of pairs.  Sorry, I am too tired to think right now.  You are right that I only worried about equivalence with this logic implementation.  Please ignore this comment.

src/main/java/net/imglib2/algorithm/fill/ComparableComparator.java
hanslovsky: I agree that using `Object.equals` makes (more) sense as we are not interested in any order. So essentially, that means `EqualsComparator` as you specified for unrestricted `T`, and `TypeComparator` for `T extends Type< T >` and Other than that, I agree on being tired and I will have a second look at it tomorrow.

src/main/java/net/imglib2/algorithm/fill/FloodFill.java
```diff
@@ -0,0 +1,207 @@
...
+     * @param <U> No restrictions on U. Appropriate comparator and writer is the only requirement.
+     */
+    public static < T, U > void fill(
+            final RandomAccessible< T > source,
+            final RandomAccessible< U > target,
+            final Localizable seed,
+            final T seedLabel,
+            final U fillLabel,
+            final Shape shape,
+            final Comparator< Pair< T, U > > comparator,
```
hanslovsky: The more I think about it, the more I think, we should not use the `Comparator` interface as it does not express what we are actually interested in. We are interested only in a true/false decision. Returning an order does not provide any useful information and makes things less intuitive. Ideally, we would have the same interface as `Comparator` that returns `Boolean` instead of `Integer`. Java 8's [BiFunction](https://docs.oracle.com/javase/8/docs/api/java/util/function/BiFunction.html) seems like a reasonable choice. Instead of `Comparator< Pair< T, U > > comparator` we would then pass `BiFunction< Pair< T, U >, Pair< T, U >, Boolean > checkLocation` (or better name if we find one). Using this interface would conform more with what we actually use the current `comparator` for.

hanslovsky:
Replaced Comparator by Filter interface, which makes most of the previous discussion obsolete.
The merge still requires and will not happen before the release of imglib2-2.9.0 to resolve the [SNAPSHOT dependency issue](https://github.com/imglib/imglib2-algorithm/pull/22#discussion_r60838078). Also [here](https://github.com/imglib/imglib2-algorithm/pull/22#discussion_r60838084).

ctrueden:
@tpietzsch already [released `2.9.0`](https://github.com/imglib/imglib2/releases/tag/imglib2-2.9.0) earlier today.

hanslovsky:
Thanks for the hint, @ctrueden, and the release, @tpietzsch. Pull request should be ready to be merged now.

axtimwalde:
Thanks @hanslovsky

ctrueden:
Thanks!

src/main/java/net/imglib2/algorithm/morphology/table2d/Spur.java
```diff
@@ -73,7 +73,7 @@
 
 		final ImgFactory< T > factory;
 		if ( extendedVal instanceof NativeType )
-			factory = ( ImgFactory< T > ) Util.getArrayOrCellImgFactory( target, ( NativeType ) extendedVal );
+			factory = Util.< NativeType >getArrayOrCellImgFactory( target, ( NativeType ) extendedVal );
```
tpietzsch: @hanslovsky @axtimwalde For me this is a compile error in Eclipse. Strangely it is not for mvn on the command line. In this case, I have to say, I agree with Eclipse: `ImgFactory<NativeType> != ImgFactory<T>`. I wonder why this compiles. Which Eclipse version are you using?

hanslovsky: I am using IntelliJ IDEA 2016.1.1, java 1.8.0_92 (previously 1.8.0_77), maven 3.3.9. Both compiling inside IDEA and running `mvn clean package` works. What's your java version?

axtimwalde: @tpietzsch @hanslovsky both java8 and eclipse are stupid, I can infer what's going on and eventually it doesn't make a difference anyways.  But who am I to judge compiler design ;).  @hanslovsky's version would save an explicit runtime cast but eclipse not accepting it is a red flag, so I reverted it:

https://github.com/imglib/imglib2-algorithm/commit/b7bdb933ea42cc1fb233e3006f341f05c5897824

hanslovsky: As a side note: Closing IDEA and re-opening the project would trigger the same error. "Re-writing" to the exact same code would make the error go away.

tpietzsch: Insane! But I can easily top that with Eclipse:
If you have
```
( ImgFactory< T > ) Util.getArrayOrCellImgFactory( target, ( NativeType ) extendedVal );
```
and tell Eclipse to "remove unnecessary casts", it will remove the "unnecessary cast".
```
Util.getArrayOrCellImgFactory( target, ( NativeType ) extendedVal );
```
then complain with a compile error...

PR #26, opened 2016-06-28
Grid added

SimonSchmid:
allows to create a grid of patches of an image

### DIFF
--- a/src/main/java/net/imglib2/algorithm/grid/GridRandomAccess.java
+++ b/src/main/java/net/imglib2/algorithm/grid/GridRandomAccess.java
@@ -1,0 +1,213 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2015 Tobias Pietzsch, Stephan Preibisch, Barry DeZonia,
+ * Stephan Saalfeld, Curtis Rueden, Albert Cardona, Christian Dietz, Jean-Yves
+ * Tinevez, Johannes Schindelin, Jonathan Hale, Lee Kamentsky, Larry Lindsey, Mark
+ * Hiner, Michael Zinsmaier, Martin Horn, Grant Harris, Aivar Grislis, John
+ * Bogovic, Steffen Jaensch, Stefan Helfrich, Jan Funke, Nick Perry, Mark Longair,
+ * Melissa Linkert and Dimiter Prodanov.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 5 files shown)


src/main/java/net/imglib2/algorithm/grid/Patch.java
```diff
@@ -0,0 +1,278 @@
...
+		if (!Intervals.contains(srcImage, raiPos)) {
+			throw new IndexOutOfBoundsException("IndexOutOfBoundsException");
+		}
+		RandomAccessibleInterval<T> raiPatch = Views.interval(srcImage, min, max);
+		// extend the source image if the patch and borders overlap
+		if (!Intervals.contains(srcImage, raiPatch)) {
+			raiPatch = Views.interval(Views.extendMirrorSingle(srcImage), min, max);
+		}
+
+		this.patch = Views.zeroMin(raiPatch);
```
axtimwalde: A new `View` is generated whenever `Patch` is returned by `GridRandomAccess.get()`.  This makes the `Patch` class not only obsolete but redundant.  Try to implement a solution that returns `Views.offsetInterval(long[], long[]);` directly in `GridRandomAccess.get()`.

src/main/java/net/imglib2/algorithm/grid/GridRandomAccess.java
```diff
@@ -0,0 +1,213 @@
...
+import net.imglib2.RandomAccess;
+import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.Sampler;
+
+/**
+ * {@link RandomAccess} for a grid.
+ * 
+ * @author Simon Schmid (University of Konstanz)
+ */
+public class GridRandomAccess<T> implements RandomAccess<RandomAccessibleInterval<T>> {
```
axtimwalde: Extend net.imglib2.Point to remove all `Localizable` and `Positionable` cruft.

axtimwalde:
@SimonSchmid thanks for the initiative.  Could you please check if what you are trying here isn't already possible with a combination of [`RectangleShape`](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/neighborhood/RectangleShape.java#L73) or [`CenteredRectangleShape`](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/neighborhood/CenteredRectangleShape.java#L89) in combination with [`SubsampleIntervalView`](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/view/Views.java#L907) and some coordinate shifting with [`Views.offsetInterval(...)`](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/view/Views.java#L676)?

src/main/java/net/imglib2/algorithm/grid/Patch.java
SimonSchmid: I didn't know about this method, thanks!

src/main/java/net/imglib2/algorithm/grid/GridRandomAccess.java
SimonSchmid: Nice hint!

SimonSchmid:
@axtimwalde Thanks for the comments. I'll try your proposed solution, that could work.

dietzc:
Keep in mind that `RectangleShape` neighborhood are not `RandomAccessible`!

axtimwalde:
@dietzc true, but shouldn't it be `RandomAccessible`?

dietzc:
@axtimwalde yes, even if I don't know if `Shape`s in general should be or just special cases? Because, for cases where the `Shape` is not a complete `Interval` (so everything besides `RectangleShape`) you would have to have e.g. a `RandomAccessible<Pair<BoolType,T>>`. The `BoolType` can be used for inside-outside testing. But I haven't really thought this through. Maybe @tpietzsch can provide more insight?

axtimwalde:
For now I suggest special case, independent of the Shape interface because Rectangles are intervals.  @dietzc & @SimonSchmid, are you fine with me closing this pull request?  This is too far from ready and should be discussed in an issue or elsewhere.

dietzc:
You can close it. We will keep the code on our side until it's clear if `RectangleShape`s can be `RandomAccessible` or not. I will discuss this with @tpietzsch at the upcoming hackathon.

PR #30, opened 2016-11-28
bugfix: randomAccess() without interval from NeighborhoodsAccessible …

tpietzsch:
…should add span to underlying source interval (if available).

This fixes https://github.com/imglib/imglib2/issues/144


tpietzsch:
This PR fixes the problem https://github.com/imglib/imglib2/issues/144, but it is just a patch...
This is once more an indication that the behaviour of the `randomAccess()` method without arguments should be changed from the current behaviour (described here:
https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/RandomAccessible.java#L59-L104
http://javadoc.imagej.net/ImgLib2/net/imglib2/RandomAccessible.html#randomAccess--
) to cover as much extent as possible (not restricted to the domain of an RAI).

PR #31, opened 2016-11-29
Add RandomAccessiblePair randomAccess test

stelfrich:
Adds a test that fails prior to #30.

### DIFF
--- a/src/test/java/net/imglib2/algorithm/neighborhood/RandomAccessiblePairWithNeighborhoodTest.java
+++ b/src/test/java/net/imglib2/algorithm/neighborhood/RandomAccessiblePairWithNeighborhoodTest.java
@@ -1,0 +1,48 @@
+package net.imglib2.algorithm.neighborhood;
+
+import net.imglib2.Cursor;
+import net.imglib2.FinalDimensions;
+import net.imglib2.RandomAccess;
+import net.imglib2.RandomAccessible;
+import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.img.Img;
+import net.imglib2.img.array.ArrayImgFactory;
+import net.imglib2.outofbounds.OutOfBoundsBorderFactory;
+import net.imglib2.outofbounds.OutOfBoundsFactory;
+import net.imglib2.type.numeric.integer.ByteType;
+import net.imglib2.util.Pair;
+import net.imglib2.view.Views;
+
+import org.junit.Test;
+
+public class RandomAccessiblePairWithNeighborhoodTest {
+
+	@Test
+	public void test() {
+		// Extend input with Views.extend()
+		Img<ByteType> in = new ArrayImgFactory< ByteType >().create( new FinalDimensions(3, 3), new ByteType() );
+		OutOfBoundsFactory<ByteType, RandomAccessibleInterval<ByteType>> outOfBoundsFactory =
+			new OutOfBoundsBorderFactory<>();
+		RandomAccessibleInterval<ByteType> extendedIn = Views.interval(Views.extend(in, outOfBoundsFactory), in); // FAILS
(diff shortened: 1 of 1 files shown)


src/test/java/net/imglib2/algorithm/neighborhood/RandomAccessiblePairWithNeighborhoodTest.java
```diff
@@ -0,0 +1,82 @@
...
+{
+
+	@Test
+	public void testRandomAccessOnPair()
+	{
+		// Extend input with Views.extend()
+		Img< ByteType > in = new ArrayImgFactory< ByteType >().create( new FinalDimensions( 3, 3 ), new ByteType() );
+		OutOfBoundsFactory< ByteType, RandomAccessibleInterval< ByteType > > outOfBoundsFactory =
+				new OutOfBoundsBorderFactory<>();
+		RandomAccessibleInterval< ByteType > extendedIn = Views.interval( Views.extend( in, outOfBoundsFactory ), in );
```
hanslovsky: Please use `Views.extendBorder` instead of creating a separate factory.

src/test/java/net/imglib2/algorithm/neighborhood/RandomAccessiblePairWithNeighborhoodTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+		// Create RandomAccessiblePair
+		final RandomAccessible< Neighborhood< ByteType > > safe = new RectangleShape( 1,
```
hanslovsky: ` 1,` should be in a new line like this (with appropriate indentation):
```java
... = new RectangleShape( 
				1,
				false )
```

src/test/java/net/imglib2/algorithm/neighborhood/RandomAccessiblePairWithNeighborhoodTest.java
```diff
@@ -0,0 +1,82 @@
...
+				false ).neighborhoodsRandomAccessibleSafe( extendedIn );
+		RandomAccessible< Pair< Neighborhood< ByteType >, ByteType > > pair = Views.pair( safe, extendedIn );
+
+		// Set position out of bounds
+		RandomAccess< Pair< Neighborhood< ByteType >, ByteType > > randomAccess = pair
```
hanslovsky: No need for line break here (there are longer lines in this file).

src/test/java/net/imglib2/algorithm/neighborhood/RandomAccessiblePairWithNeighborhoodTest.java
stelfrich: @hanslovsky I have added an additional comment in the opening line so that the formatter doesn't try to move the `1` to the first line every time it's run. Let me know if I should remove it or keep it that way...

stelfrich:
Thanks for looking at this @hanslovsky! I have added a comment to one of yours for your review.

src/test/java/net/imglib2/algorithm/neighborhood/RandomAccessiblePairWithNeighborhoodTest.java
hanslovsky: That's interesting. Are you using eclipse? I have noticed that the eclipse formatter is not always consistent among different machines (probably dependent on what settings are being used). I think the comment is fine here.

stelfrich: I was using Eclipse Neon on macOS with a weird configuration (that I don't have a clue where I got it from) instead of the `imglib2-eclipse-mars-clean-up-formatter-style.xml`. Sorry for the confusion!

PR #32, opened 2017-02-24
Add Hough transform

panovr:
This pull request is to port the Hough transform from ImgLib1 to ImgLib2, and also see this issue: https://github.com/imagej/imagej-ops/issues/465


hanslovsky:
Just as a note: The link to issue 465 is broken in your comment. I can c&p the text though.

panovr:
@hanslovsky fixed

pom.xml
```diff
@@ -37,13 +37,27 @@
 			<groupId>net.imglib2</groupId>
 			<artifactId>imglib2-realtransform</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>net.imglib2</groupId>
+			<artifactId>imglib2-algorithm-gpl</artifactId>
```
ctrueden: This is a problem. The `imglib2-algorithm-gpl` component must not be a dependency here, both for legal reasons, as well as logistical/organizational ones.

pom.xml
```diff
@@ -37,13 +37,27 @@
...
+		</dependency>
+		<dependency>
+			<groupId>net.imglib2</groupId>
+			<artifactId>imglib2-ij</artifactId>
```
ctrueden: This is a problem. We must not depend on ImageJ 1.x here.

pom.xml
```diff
@@ -37,13 +37,27 @@
...
 		<!-- ImageJ dependencies -->
 		<dependency>
 			<groupId>net.imagej</groupId>
 			<artifactId>imagej-common</artifactId>
 		</dependency>
 
+		<!-- SCIFIO dependencies -->
+		<dependency>
+			<groupId>io.scif</groupId>
+			<artifactId>scifio</artifactId>
```
ctrueden: We must not depend on SCIFIO here. The `imglib2-algorithm` component should not need to read/write image formats.

.gitignore
```diff
@@ -2,3 +2,7 @@
 /.project
 /.settings/
 /target/
+
+# IntelliJ
```
ctrueden: This should be a separate commit, with message like "Ignore IntelliJ metadata files"

src/main/java/net/imglib2/algorithm/hough/HoughLineTransform.java
```diff
@@ -0,0 +1,292 @@
...
+ *        | y |         | -sin(theta) |       | cos(theta) |
+ *
+ * In other words, rho represents the signed minimum distance from the image origin to the line,
+ * and theta indicates the angle between the row-axis and the minimum offset vector.
+ *
+ * For a given point, then, votes are placed along the curve
+ *
+ * rho = y * sin(theta) + x * cos(theta)
+ * @Override
+ *
```
ctrueden: Add yourself, as well as the original author(s), as `@author`s here.

src/main/java/net/imglib2/algorithm/hough/HoughLineTransform.java
```diff
@@ -0,0 +1,292 @@
...
+ * l(t) = | x | = rho * |  cos(theta) | + t * | sin(theta) |
+ *        | y |         | -sin(theta) |       | cos(theta) |
+ *
+ * In other words, rho represents the signed minimum distance from the image origin to the line,
+ * and theta indicates the angle between the row-axis and the minimum offset vector.
+ *
+ * For a given point, then, votes are placed along the curve
+ *
+ * rho = y * sin(theta) + x * cos(theta)
+ * @Override
```
ctrueden: The `@Override` annotation makes no sense here.

src/main/java/net/imglib2/algorithm/hough/HoughLineTransform.java
```diff
@@ -0,0 +1,292 @@
...
+    public final double dTheta;
+    public final double dRho;
+    private final T threshold;
+    private final int nRho;
+    private final int nTheta;
+    private final double[] rho;
+    private final double[] theta;
+    private ArrayList<double[]> rtPeaks;
+
+    final private static float computeLength(final long[] position) {
```
ctrueden: Please apply the ImgLib2 [style formatting](https://github.com/imglib/imglib2/tree/master/doc) to this class, so formatting is consistent with the rest of ImgLib2.

ctrueden:
Thanks so much for working on this! It is looking really, really good!

Three main concerns:

1. Extra dependencies are undesirable. I did not see where you actually need them? Probably they can just be removed?
2. Style formatting would be nice to match with the rest of ImgLib2. There are Eclipse formatting rules for this, but maybe you want to make IntelliJ version of them and commit it to ImgLib2 repository, to help others who use IntelliJ?
3. We need unit tests.

If you can address the above three things, I am happy.

tinevez:
I notice that this PR covers hough line transform. Are there plans for hough circle transform?

panovr:
@ctrueden I really appreciate for the detailed code review, thanks!

FIX:
- the undesirable dependencies exist because I write a test class for validating. I should delete these.
-  the `imglib2-algorithm-gpl` is need for the `PickImagePeaks` class in HoughTransform.java at [this line](https://github.com/imglib/imglib2-algorithm/pull/32/files#diff-37e3a5f646c482366589a37aef2cdda0R158)
- I format code using default IntelliJ code format, and I will re-format code to ImgLib2 style.
- I will complement unit tests.

TODO:
- Add a method for generate lines coordinates based on vote images
- Add hough circle transform or need others contribution @tinevez 

HELP:
- I still don't know how to use the method `public void setSuppression( final double[] r )` in  class `PickImagePeaks.java` of file [PickImagePeaks](https://github.com/imglib/imglib2-algorithm-gpl/blob/master/src/main/java/net/imglib2/algorithm/math/PickImagePeaks.java) after the vote space image has been computed if I want to invoke the method `public ArrayList< long[] > getPeakList()` at [this line](https://github.com/imglib/imglib2-algorithm-gpl/blob/master/src/main/java/net/imglib2/algorithm/math/PickImagePeaks.java#L304) @tinevez

ctrueden:
> the imglib2-algorithm-gpl is need for the PickImagePeaks class

OK. If the implementation would be more complicated without using this class, then I propose to add these Hough classes instead to imglib2-algorithm-gpl, to avoid circular dependency, and to have correct, legal licensing.

> I will complement unit tests. Any example I can follow?

There are tons of unit tests across many components. We use JUnit. You could look at e.g. `DilationTest` in this repository to see how Jean-Yves tests the morphological dilate operation.

hanslovsky:
What exactly is the `PickImagePeaks` class used for in this case? Is it simply to find local extrema? You could use [`LocalExtrema`](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java) instead.

ctrueden:
@panovr Did you try with `LocalExtrema` yet?

panovr:
@hanslovsky @ctrueden I will investigate `LocalExtrema` later.

tinevez:
I am taking care of the Hough circle transform.

ctrueden:
@tinevez OK, thank you very much!

imagejan:
@tinevez are you planning to migrate your work in https://github.com/tinevez/CircleSkinner/tree/master/src/main/java/net/imagej/circleskinner/hough to `imglib2-algorithm`?

tinevez:
Yes, before end of June.

ctrueden:
@tinevez I have a new undergrad student, @gselzer, working on ImageJ Ops and ImgLib2 stuff now. If there is anything he could take off your plate relating to the Hough transform project, just let us know!

tinevez:
Christ I am late everywhere. 
@gselzer can we chat over gitter sometime about this?

gselzer:
@tinevez I'm sorry that I did not see this comment before, but I am just now beginning to move onto this project and would love to chat on Gitter!

ctrueden:
See also imagej/imagej-ops#526.

hanslovsky:
@panovr did you have a look at `LocalExtrema` and whether or not that could (partially) replace `PickImagePeaks`?

ctrueden:
@hanslovsky FYI, @panovr has moved on from LOCI (back to China, where has main project is machine learning classification and segmentation of bird species). But our student @gselzer could (will! :wink:) work on completing this PR.

hanslovsky:
Thanks for the update @ctrueden

gselzer:
@hanslovsky @ctrueden Fixes:
* use `LocalExtrema` instead of `PickImagePeaks`
* format to imglib standards

All commits build and all tests pass on each commit

ctrueden:
@hanslovsky @tpietzsch After going through this PR with @gselzer and cleaning things up, we are now reasonably satisfied with it. The dependency on GPL code was purged.

It is questionable whether various calculation methods should really be returning `boolean` to indicate success. What do you think? Any further changes you'd like to see before merging this?

hanslovsky:
@ctrueden @gselzer I haven't had time to go through the code yet but hopefully will be able do so this week. I agree that reporting success through `boolean` return value is probably not necessary but more once I had a closer look.

src/main/java/net/imglib2/algorithm/hough/HoughTransform.java
```diff
@@ -0,0 +1,258 @@
...
+		peaks = new ArrayList();
+		long[] dims = new long[ image.numDimensions() ];
+
+		for ( Point p : peaksList )
+		{
+			for ( int i = 0; i < p.numDimensions(); i++ )
+			{
+				dims[ i ] = p.getLongPosition( i );
+			}
+			peaks.add( dims );
```
hanslovsky: Bug, `dims` is not cloned before adding -- all peaks are the same.

src/main/java/net/imglib2/algorithm/hough/HoughTransform.java
```diff
@@ -0,0 +1,258 @@
...
+	 *
+	 * @return whether peak picking was successful
+	 */
+	protected boolean pickPeaks( ExecutorService exService )
+	{
+		S maxValue = voteSpace.firstElement().copy();
+		maxValue.setReal( maxValue.getRealDouble() );
+		final MaximumCheck< S > maxCheck = new MaximumCheck< S >( maxValue );
+
+		List< Point > peaksList = LocalExtrema.findLocalExtrema( voteSpace, maxCheck, exService );
```
hanslovsky: Avoid use of deprecated methods.

src/main/java/net/imglib2/algorithm/hough/HoughTransform.java
```diff
@@ -0,0 +1,258 @@
...
+		peaks = new ArrayList();
```
hanslovsky: `ArrayList` needs parameters:
> ArrayList is a raw type. References to generic type ArrayList<E> should be parameterized

src/main/java/net/imglib2/algorithm/hough/HoughTransform.java
```diff
@@ -0,0 +1,258 @@
...
+
+	/**
+	 * Pick vote space peaks with a {@link LocalExtrema}.
+	 *
+	 * @return whether peak picking was successful
+	 */
+	protected boolean pickPeaks( ExecutorService exService )
+	{
+		S maxValue = voteSpace.firstElement().copy();
+		maxValue.setReal( maxValue.getRealDouble() );
```
hanslovsky: This seems like a bug. The minimal peak value should probably be zero or one. What if `voteSpace.firstElement()` is the highest peak? Then, no other peak would be found.

hanslovsky:
I finally had some time to go over the code. I feel that this PR needs some work before we can merge it. These are my (itemized) thoughts: 
 - Using stateful classes/objects seems suboptimal to me here. We are not interested in the state/do not change the state over time but run the computation once. Instead, static methods for vote space generation should be implemented. Ideally, split up vote space generation and maxima localization. Not using stateful classes/objects would also avoid variables not being initialized inside the constructor (most of the times, [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) is desirable).
 - `RandomAccessibleInterval` should be used as parameter over `Img`.
 - Instead of passing an image factory to allocate vote space, pass a `RandomAccessibleInterval` allocated by the caller. That way, `OutputAlgorithm.getResult` becomes obsolete for this class and static methods can be used as suggested above.
 - This class is not a benchmark and therefore should not implement `Benchmark`
 - final variables should be declared final
 - peaks in the votespace should not be `List< long[] >` but `List< Point >` instead:
   - Converting to `List< long[] >` creates unnecessary copies, and
   - introduces a bug because [the `long[]` array is not cloned before adding, i.e. all maxima are the same](https://github.com/imglib/imglib2-algorithm/pull/32/files#diff-37e3a5f646c482366589a37aef2cdda0R218).

   This will make `HoughTransform.pickPeaks` trivial/obsolete. As there are several bugs in this method (cf comments in diff), this seems like a good idea.
 - Do not use any parallelization for the basic implementation, i.e. do not require any `ExecutorService` or related classes/methods to be passed. Parallelization can be implemented on top of single-threaded basic functionality (if desired at all).
 - Avoid the use of deprecated classes, in particular `findLocalExtrema( RandomAccessibleInterval, LocalNeighborhoodCheck, ExecutorService )`. This is the only method that requires `ExecutorService` and should be replaced by one of the more recent `LocalExtrema.findLocalExtrema` signatures that do not provide parallelization.

ctrueden:
Thanks @hanslovsky! @gselzer Please make the changes when you get a chance.

src/main/java/net/imglib2/algorithm/hough/HoughTransform.java
gselzer: Yes, you are right. I do not know how I missed this. It is fixed now.

src/main/java/net/imglib2/algorithm/hough/HoughTransform.java
gselzer: I switched this to a version that no longer uses an `ExecutorService`, and removed the service from the program. See commit [83b6b25](https://github.com/imglib/imglib2-algorithm/pull/32/commits/83b6b25af3e717d3cb01ed56051fc65259f621d4) for details.

src/main/java/net/imglib2/algorithm/hough/HoughTransform.java
gselzer: D'oh! You are right. Fixed now.

gselzer:
@hanslovsky I agree that the votespace peaks could be `List<Point>` instead of `List<long[]>`. I fail to understand however how changing this would make `HoughTransform.pickPeaks()` obsolete, unless you are proposing moving the peak picking to another location. I think that going forward with this would certainly trim the method down but it should not be obsolete. Can you elaborate on this?

hanslovsky:
@gselzer With `List<Point>` instead of `List<long[]>` `HoughTransform.pickPeaks` would be reduced to a single line (or two, considering that we have a generic `T` instead of `IntType`):
```
LocalExtrema.findLocalExtrema( voteSpace, new LocalExtrema.MaximumCheck<>( new IntType( 1 ) ) );
```
In general, users might want to use their own ways to detect local extrema, e.g. to minimize noise but your are right that it is probably best to leave the `pickPeaks` method in place for users who do not.

gselzer:
@hanslovsky @ctrueden I have fixed all of the comments above, either in their own commit or in the larger commit where I refactored the algorithm to use static methods instead of stateful classes. Let me know any further comments you might have. I massaged the old votespace regression test so that that now passes, and am currently writing a peak finder regression test, since such a test was never written, and the results are looking promising. The only problem that I have is that the width of the vote space is over the wrong interval; currently this interval is [0, maxRho * 2), whereas we would like [-maxRho, maxRho) to be the interval. Is there a way to change this interval?

gselzer:
My solution to the interval problem was by creating a translated `IntervalView` within `PickPeaks`, which allows us to return the correct polar coordinates for our line without a) automatically mutating the input at any point or b) requiring the user to do so before getting the peaks from the vote space. As for now I do not see any reason to make this optional, as without this translation `PickPeaks` will not return the correct polar coordinate results, however if there is a reason to make this optional I can do so.

I also added static methods to generate the slope and intercept from the polar coordinate output, as well as tests for all relevant methods.

Let me know if there are any issues.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,343 @@
...
+			imageCursor.fwd();
+			imageCursor.localize( position );
+
+			for ( int t = 0; t < nTheta; ++t )
+			{
+				if ( imageCursor.get().compareTo( threshold ) > 0 )
+				{
+					fRho = Math.cos( theta[ t ] ) * position[ 0 ] + Math.sin( theta[ t ] ) * position[ 1 ];
+					r = Math.round( ( float ) ( ( fRho - minRho ) / dRho ) );
+					voteLoc[ 0 ] = r;
```
hanslovsky: Use 
```
outputRA.setPosition( r, 0 );
outputRA.setPosition( rt 1 );
```
instead of writing into `voteLoc` and then copying into `outputRA`.
This will also make `voteLoc` obsolete (no need to create an int[] for every pixel of `input`)

src/test/java/net/imglib2/algorithm/hough/HoughLineTransformTest.java
```diff
@@ -0,0 +1,171 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+
+package net.imglib2.algorithm.hough;
+
+import static org.junit.Assert.assertEquals;
+
+import java.awt.image.BufferedImage;
```
hanslovsky: Is image IO really necessary here? imglib2-algorithm should be agnostic of any ui libraries, even within test cases. A hand-designed ground truth example would be preferable, ideally with differently sized dimensions.

src/test/java/net/imglib2/algorithm/hough/HoughLineTransformTest.java
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	@Test
+	public void testEquationConversion()
+	{
+		final int rho = 2, theta = 45;
+		final double slope = HoughTransforms.getSlope( theta );
+		final double yint = HoughTransforms.getIntercept( rho, theta );
+
+		assertEquals( -1.0000000000000002, slope, 0 );
```
hanslovsky: Instead of using a high-precision floating point value here, pass a small delta, e.g. `1e-8` to `assertEquals`:
```
Assert.assertEquals( 1.0, slope, 1e-8 );
```
Please do the same for the assert on `yint`

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,343 @@
...
+
+	/**
+	 * Calculates a default number of rho bins, which corresponds to a
+	 * resolution of one pixel.
+	 *
+	 * @param inputImage
+	 *            the {@link RandomAccessibleInterval} in question.
+	 * @return default number of rho bins.
+	 */
+	private static int defaultRho( final RandomAccessibleInterval< ? > inputImage )
```
hanslovsky: replace `final RAI inputImage` with `final Dimensions dimensions`.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,343 @@
...
+	/**
+	 * Returns the size of the vote space output image given an input
+	 * {@link RandomAccessibleInterval}.
+	 * 
+	 * @param input
+	 *            - the {@link RandomAccessibleInterval} over which the Hough
+	 *            Line Transform will be run
+	 * @return {@code long[]} - the dimensions of the vote space image
+	 */
+	public static < T extends RealType< T > > long[] getVotespaceSize( RandomAccessibleInterval< T > input )
```
hanslovsky: replace `final RAI inputImage` with `final Dimensions dimensions`, no need for generic parameter `T`.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,343 @@
...
+		return new long[] { nRho, nTheta };
+	}
+
+	/**
+	 * Pick vote space peaks with a {@link LocalExtrema}.
+	 *
+	 * @return {@code List<Point>} - a list of all of the local maxima of the
+	 *         {@code voteSpace}.
+	 */
+	public static < T extends RealType< T > & NativeType< T > > List< Point > pickPeaks( RandomAccessibleInterval< T > voteSpace, T minPeak )
```
hanslovsky: Line is too long, break it like this:
```java
public static < T extends Comparable< T > > List< Point > pickPeaks(
			final RandomAccessibleInterval< T > voteSpace,
			final T minPeak )
```

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,343 @@
...
+	 * parameter with the results.
+	 * 
+	 * @param input
+	 *            - the {@link RandomAccessibleInterval} to run the Hough Line
+	 *            Transform over
+	 * @param votespace
+	 *            - the {@link RandomAccessibleInterval} in which the results
+	 *            are stored
+	 */
+	public static < T extends RealType< T > > void voteLines( final RandomAccessibleInterval< T > input, final RandomAccessibleInterval< T > votespace )
```
hanslovsky: Line is too long, break it like this:
```java
public static < T extends RealType< T > > void voteLines(
			final RandomAccessibleInterval< T > input,
			final RandomAccessibleInterval< T > votespace )
```

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,343 @@
...
+	 *            Transform over
+	 * @param votespace
+	 *            - the {@link RandomAccessibleInterval} in which the results
+	 *            are stored
+	 * @param nRho
+	 *            - the number of bins for rho resolution
+	 * @param nTheta
+	 *            - the number of bins for theta resolution
+	 */
+	public static < T extends RealType< T > > void voteLines( final RandomAccessibleInterval< T > input, final RandomAccessibleInterval< T > votespace, final int nRho, final int nTheta )
```
hanslovsky: Line is too long, break it like this:
```java
public static < T extends RealType< T > > void voteLines( 
			final RandomAccessibleInterval< T > input,
			final RandomAccessibleInterval< T > votespace, 
			final int nRho,
			final int nTheta )
```

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,343 @@
...
+	 * @param nRho
+	 *            - the number of bins for rho resolution
+	 * @param nTheta
+	 *            - the number of bins for theta resolution
+	 * @param threshold
+	 *            - the minimum value allowed by the populator. Any input less
+	 *            than or equal to this value will be disregarded by the
+	 *            populator.
+	 */
+	public static < T extends RealType< T > > void voteLines( final RandomAccessibleInterval< T > input, final RandomAccessibleInterval< T > votespace, final int nRho, final int nTheta, final T threshold )
```
hanslovsky: Line is too long, break it like this:
```java
public static < T extends RealType< T > > void voteLines( 
			final RandomAccessibleInterval< T > input, 
			final RandomAccessibleInterval< T > votespace,
			final int nRho, 
			final int nTheta,
			final T threshold )
```

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
hanslovsky: Add generic parameter `U extends RealType< U >` for `votespace` parameter

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
hanslovsky: Add generic parameter `U extends RealType< U >` for `votespace` parameter

src/test/java/net/imglib2/algorithm/hough/HoughLineTransformTest.java
gselzer: @hanslovsky I have addressed all of the issues before this one in the list, and will continue with the remaining issues next week.

hanslovsky: @gselzer I just went through the additional commits, looks good so far!

ctrueden:
@hanslovsky Regarding your comment about avoiding `ImageIO` in favor of a hand-designed ground truth: it would be helpful to have some clarification about what the concerns are. Is it the usage of a `javax` class in the test scope? Are you concerned about the test being incompatible with some JVMs? Or is there some other reason you do not like the result image saved as a resource? The alternative—to embed a large primitive array in the test class—seems no better to me, as it would be larger on disk.

hanslovsky:
@ctrueden My issues with the current test are two-fold: 
 - Using a ground-truth image hides how the image was generated. With a large image like that it seems to me the image itself was created with the algorithm it is supposed to test. A smaller, hand-crafted test-case would be more useful:
   - Out of the 501x501 input image, only a very small subset is used -> the input size can be reduced drastically
   - For a smaller input, the ground truth image size will be smaller. That size can be further reduced with a coarser choice of the parameters. With a smaller ground truth image, the ground truth image can be populated within the code (pixel-by-pixel for all non-zero pixels).
 - The test class introduces `javax` and `awt` dependencies. imglib2 is an image processing library and not an image io/ui library, and therefore should avoid dependencies on ui framework libraries. This is my personal oppinion and maybe the @imglib/admins should comment on that.

gselzer:
@hanslovsky I do not know if this changes your opinion on the matter however the size of the test image is 50 x 50, not 501 x 501 (the 50s are being specified as `long`). With that being said we could refactor this test to a rectangular input image if that would solve any issues.

hanslovsky:
Thanks for the clarification @gselzer 
I still believe that the ground truth for the test should be hand crafted within the code for reproducability. If we don't know how the test ground truth was generated we will not be able to know if the test fails because the algorithm is wrong or because we messed up something in the test.

I suggest:
 - Make the test image and (especially) the vote space ground truth even smaller (and not square) such that construction by hand is feasible
 - Create a larger test image with (at least) 2 (non-continuous) straight lines and test at a higher level that you get exactly two (or n if you use more lines) peaks with the correct parameters

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,388 @@
...
+		while ( imageCursor.hasNext() )
+		{
+			double fRho;
+			int r;
+
+			imageCursor.fwd();
+
+			for ( int t = 0; t < nTheta; ++t )
+			{
+				if ( filter.test( imageCursor.get() ) )
```
hanslovsky: `filter.test` should be outside the loop over theta. If the pixel is not selected we should not waste any time iterating over all thetas and re-evaluating the test over and over again. (Then, you can also combine `imageCursor.fwd()` and `imageCursor.get()` into
```java
if ( filter.test( imageCursor.next() )
```

gselzer:
@hanslovsky I will rewrite the tests such that we test against a smaller, non-square image with 2+ lines. I agree that this will create a better test. However I still do not understand what you mean by a "hand-crafted" ground truth. My understanding of this (and correct me if I am wrong) is that you desire a way in the code of determining what the value should be at a particular pixel instead of relying on an image. This is understandable, but because there is really no equation that describes these vote spaces the only options that I see here are:

1. storing a rather large `byte[]` within the test
2. using a rather large (though less so) hashmap.

I do not really know which of these would be the right option.

hanslovsky:
@gselzer Yes, you understood me correctly there. The reason that I would like to see this is that the current test image was probably created via the tested algorithm itself and we thus don't have any means of verifying that the algorithm actually worked correctly. While it is true that we cannot generate the ground truth via a closed form equation, the vote space is still generated in a deterministic way. I would even prefer a pixel-by-pixel (for non-trivial pixels) initialization of the vote space, something like this:
```
RandomAccess voteAccess;
voteAccess.setPosition( nonZeroPosition );
voteAccess.get().set( nonZeroCount );
```
For the current parameter setting, 2468 out of a total of 25380 pixels are non-zero. That is obviously way too much to generate by hand. If we can get the number of non-zero pixels to down to a maintainable amount by reducing the vote space size and image size, we should do that. If you think that doing this is too much work we should just abandon that test and opt for a higher level test on the peaks only (as described above). Otherwise we would just run something like this:

```java
int doSomeComputation();
assertEquals( doSomeComputation(), doSomeComputation() );
```

gselzer:
@hanslovsky 

> The reason that I would like to see this is that the current test image was probably created via the tested algorithm itself and we thus don't have any means of verifying that the algorithm actually worked correctly.

Why is it that we do not know that the algorithm did not work correctly? It is not difficult to take the output point(s) of the algorithm, convert them into the line(s) that they refer to, and check to see if it is(they are) similar to that of the input. For example, to generate the current test image pixels were set to 255 according to the equation `y = 24`. We have verified that the algorithm works correctly since the only maximum returned from this algorithm is at `(rho, theta) = (24, pi/2)`, which are the correct `(rho, theta)` coordinates for such a line. I guess I am confused as to what more is required to make sure that the algorithm is working correctly.

> While it is true that we cannot generate the ground truth via a closed form equation, the vote space is still generated in a deterministic way. I would even prefer a pixel-by-pixel (for non-trivial pixels) initialization of the vote space

I agree with this. However I do not understand the method by which you want to generate these values. Even with a, for example, 11x13 the smallest test image that still returns accurate data will have no less than thousands of pixels. This is way too much to calculate by hand.

>  If you think that doing this is too much work we should just abandon that test and opt for a higher level test on the peaks only (as described above). 

I would argue against this. The votespace includes a lot more information than just the peaks (it is important to make sure that all non-zero pixels are contained within the general flow of the space) and I think it is important to test for this too.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
gselzer: @hanslovsky My imglib2 style sheet is not letting me do this; if I attempt to break this as above and then format the signatures return to one long line. Is there an update I am missing?

hanslovsky:
@gselzer
> Why is it that we do not know that the algorithm did not work correctly? It is not difficult to take the output point(s) of the algorithm, convert them into the line(s) that they refer to, and check to see if it is(they are) similar to that of the input. For example, to generate the current test image pixels were set to 255 according to the equation y = 24. We have verified that the algorithm works correctly since the only maximum returned from this algorithm is at (rho, theta) = (24, pi/2), which are the correct (rho, theta) coordinates for such a line. 

In general I do agree and this is what I was referring to with my more *high-level* test that does not look at the vote space and only checks if the correct parameters are obtained.

However, this is not what is happening in [`testHoughLineTransformToTarget`](https://github.com/imglib/imglib2-algorithm/pull/32/files#diff-c86029e772453c0b53413f4ebea7d8d7R123). There, the check happens on the vote space, and the vote space is generated from `result8.png`. It is unclear how `result8.png` was generated and I can only assume it was generated by the same algorithm it is supposed to test. As stated above, testing an algorithm with a result that was produced by the same algorithm does not make any sense.

> I agree with this. However I do not understand the method by which you want to generate these values. Even with a, for example, 11x13 the smallest test image that still returns accurate data will have no less than thousands of pixels. This is way too much to calculate by hand.

For testing the vote space we are not worried about an accurate re-construction of the parameters but only that vote space gets populated correctly given a parameter precision/bin size (that is the only task of `voteLines`). Pick a small enough vote space via [`voteLines( RandomAccessibleInterval< T >, RandomAccessibleInterval< U >,  int, int, T )`](https://github.com/imglib/imglib2-algorithm/pull/32/files#diff-80a7ec57bd214e7450b0d3a56e8a5970R267) so you can populate a ground truth example by hand. Yes, that would require that you calculate the counts for all non-zero bins by hand. If that is too much work (which I can understand), drop the test for the vote space count. The higher-level check for correct reconstruction of the lines will have to suffice.

src/test/java/net/imglib2/algorithm/hough/HoughLineTransformTest.java
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	@Test
+	public < T extends RealType< T > > void testHoughLineTransformToTarget() throws IOException
+	{
+		// create votespace
+		final long[] dims = new long[ tpImg.numDimensions() ];
+		tpImg.dimensions( dims );
+		final long[] outputDims = HoughTransforms.getVotespaceSize( tpImg );
+		final Img< UnsignedByteType > votespace = new ArrayImgFactory().create( outputDims, tpImg.firstElement() );
```
hanslovsky: Create `byte[] store` of appropriate size and replace `new ArrayImgFactory().create` by `ArrayImgs.unsignedBytes( store, outputDims )`. Then, you can use `Assert.assertArrayEquals( expected, store )` instead of iterating over pixels.

src/test/java/net/imglib2/algorithm/hough/HoughLineTransformTest.java
```diff
@@ -0,0 +1,185 @@
...
+	{
+		// three expected peaks
+		final long[] expected = { -5, -89, -5, -88, 0, -76, 0, -74, 0, -73, 5, -2, 5, -1, 5, 0, 5, 1, 5, 2, 5, 88 };
+		int index = 0;
+
+		// create votespace
+		final long[] dims = new long[ tpImg.numDimensions() ];
+		tpImg.dimensions( dims );
+		final long[] outputDims = HoughTransforms.getVotespaceSize( tpImg );
+		final Img< UnsignedByteType > votespace = new ArrayImgFactory().create( outputDims, tpImg.firstElement() );
```
hanslovsky: Use `ArrayImgs.unsignedBytes( outputDims )` instead of `new ArrayImgFactory().create`

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
gselzer: Other than the first 3 my save actions match. I am also using the same style-sheet as above.

hanslovsky: I have experienced weird formatting behavior with eclipse, so that might just be it. My eclipse version is:
```
Version: Oxygen.3a Release (4.7.3a)
Build id: 20180405-1200
```
What happens if you format it, save, and after save actions were applied, you hit ctrl+z?

gselzer: Strangely enough I tried it and although it is not automatically formatting the lines it does not revert them to one long line when I format again after breaking the lines. So I guess this works.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,343 @@
...
+			theta[ t ] = dTheta * t + minTheta;
+		}
+
+		// create rho LUT
+		for ( int r = 0; r < nRho; ++r )
+		{
+			rho[ r ] = dRho * r + minRho;
+		}
+
+		final Cursor< T > imageCursor = Views.iterable( input ).localizingCursor();
```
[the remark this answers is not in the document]
gselzer: For the time being I think this is the way to go. I am still wondering if it is possible to input an image with `min != [0, ... , 0]`, but all of my attempts so far to get results from such an image have failed. I may attempt to solve this in the end but for now I will do this. My only concern was that if the user passes through a `RAI` with `min != [0, ... , 0]` the resulting `(rho, theta)` coordinates would not be technically correct (since we are translating the image we are translating the `rho` of the lines). I may attempt to fix this but with the current algorithm it may be a big time commitment

hanslovsky:
> On the contrary: such regression tests are vital to ensure that the
algorithm continues to work reproducibly in the future. It does not
guarantee correctness (but then again: does anything?), but it does verify
consistency.

Only if we can comprehend how the ground truth was generated, otherwise a future error will be rather useless. I still don't see how an appropriate example cannot be hand crafted.

gselzer:
> In general I do agree and this is what I was referring to with my more high-level test that does not look at the vote space and only checks if the correct parameters are obtained.

Ah okay now I understand what you mean by a high-level test. This is what `testPickPeaks` currently does; it takes the same image used in `testHoughLineTransformToTarget`, runs the transform and then `pickLinePeaks` on the resulting votespace, and then compares the resulting `Point`s to what is expected. 

> Only if we can comprehend how the ground truth was generated, otherwise a future error will be rather useless. I still don't see how an appropriate example cannot be hand crafted.

I think that the point @ctrueden and I are trying to make here is that this test has less to do with assuring the correctness of the algorithm and more to do with guaranteeing reproducability. This test will fail if any math changes and as such a success from this test is a sign that an output after an update to the algorithm is the same output that would have been given before. With that being said you are right in that it is not an indicator that the algorithm is working correctly; that is what `testPickPeaks` is designed to do. We can change the name of this test if that would help make it more clear that reproducability is the goal.

hanslovsky:
@gselzer I still think that a hand-crafted test would do the same thing and, on top of that, would add documentation of what is being tested. But @ctrueden is an imglib2 admin and should have the last word on this.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,446 @@
...
+		final RandomAccess< U > outputRA = votespace.randomAccess();
+
+		while ( imageCursor.hasNext() )
+		{
+			double fRho;
```
hanslovsky: No need to declare `fRho` and `r` outside the loop. Put declaration at initialization (line 393) and change `r` to `long`.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,446 @@
...
+			int r;
+
+			imageCursor.fwd();
+
+			if ( filter.test( imageCursor.get() ) )
+			{
+				for ( int t = 0; t < nTheta; ++t )
+				{
+					fRho = Math.cos( theta[ t ] ) * imageCursor.getDoublePosition( 0 ) + Math.sin( theta[ t ] ) * imageCursor.getDoublePosition( 1 );
+					r = Math.round( ( float ) ( ( fRho - minRho ) / dRho ) );
```
hanslovsky: - Do not cast to `float` before rounding. 
- Is rounding correct? That would mean that the value of each `r` bin defines the center of said bin, not the min. If that is the desired behavior, ignore this second bullet point.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,446 @@
...
+			double fRho;
+			int r;
+
+			imageCursor.fwd();
+
+			if ( filter.test( imageCursor.get() ) )
+			{
+				for ( int t = 0; t < nTheta; ++t )
+				{
+					fRho = Math.cos( theta[ t ] ) * imageCursor.getDoublePosition( 0 ) + Math.sin( theta[ t ] ) * imageCursor.getDoublePosition( 1 );
```
hanslovsky: `x` and `y` should be cached outside loop to avoid potential virtual method call overhead:
```java
if ( filter.test( imageCursor.get() ) )
{
	double x = imageCursor.getDoublePosition( 0 );
	double y = imageCursor.getDoublePosition( 1 );
	for () // continue as before.
}
```

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -0,0 +1,446 @@
...
+		final long[] dims = new long[ input.numDimensions() ];
+		input.dimensions( dims );
+
+		final double minRho = -computeLength( dims );
+		final double dRho = 2 * computeLength( dims ) / nRho;
+
+		final double minTheta = -Math.PI / 2;
+		final double dTheta = Math.PI / nTheta;
+
+		final double[] rho = new double[ nRho ];
```
hanslovsky: - `double[] rho` is never used, can be removed
 - `double[] theta` caches theta but the expensive caclulations are `Math.sin( theta )` and `Math.cos( theta )`. Cache these instead.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
gselzer: @hanslovsky you are right about the cast. I do however think that this rounding is correct since we are using `r` to approximate our value of `rho`, meaning that we want to round up in the case that `rho` is closer to the `r` greater than it than it is to the `r` less than it, if that makes sense.

hanslovsky:
The build currently fails with JavaDoc errors. Most of those errors are unrelated and fixed in master. I locally merged with master and the only remaining JavaDoc error is this:
```
[ERROR] /path/to/imglib2-algorithm/src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java:325: error: reference not found
[ERROR]          * vote space as well. Thus if {@link HoughTransforms#pickPeaks} is not used
```home/phil/workspace/bigcat-future
This is an easy fix, just repleace `pickPeaks` with `pickLinePeaks`. To confirm that JavaDoc generation does not fail you can run `mvn javadoc:javadoc` (warnings can be ignored).

File: src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java [hunk 29]
```diff
@@ -0,0 +1,446 @@
...
+import net.imglib2.algorithm.localextrema.LocalExtrema.MaximumCheck;
+import net.imglib2.type.numeric.IntegerType;
+import net.imglib2.type.numeric.RealType;
+import net.imglib2.util.Intervals;
+import net.imglib2.util.Util;
+import net.imglib2.view.IntervalView;
+import net.imglib2.view.Views;
+
+/**
+ * This abstract class provides some basic functionality for use with arbitrary
```
hanslovsky: This documentation is not accurate anymore, should probably be something along the lines of:
*This class provides static methods for the detection of lines in two-dimensional images via the Hough transform.*

hanslovsky:
@gselzer Once you update the JavaDoc to fix the error and my last requested change, this looks good for a merge to me. Yours and @ctrueden's argument regarding the test case makes sense and I agree that we should leave it as is now.

gselzer:
@hanslovsky I just fixed the error and the requested change in the most recent commit. Thanks for taking all the time and effort to get this project ready to merge!

hanslovsky:
@gselzer I looked over it once more and it seems good to me. Thank you for putting the work in and bearing with me and all my pedantic requests. I will leave this open for a little bit longer (probably until the end of the week) to see if anyone else raises concerns, and then I will merge.

Good work!

hanslovsky:
@imglib/admins What is the appropriate merge strategy? Squash and merge?

ctrueden:
@hanslovsky I personally prefer "Create a merge commit" to preserve the history, assuming each commit compiles with passing tests. Otherwise, what is the point of making clean topic branches? That said, I am not an ImgLib admin—see [Governance summary](https://imagej.net/Governance#SciJava_project_summary)—so ultimately it is up to @tpietzsch and @axtimwalde and @StephanPreibisch.

hanslovsky:
@ctrueden I was advised for my own (single-feature addition) PRs to rebase before a merge but I do not 
want to make this decision here.

ctrueden:
> I was advised for my own (single-feature addition) PRs to rebase before a merge

Doing that will do a fast-forward merge, which loses the fact that it was ever a branch. This is a bad thing which has caused social confusion and resentment in the past, where fingers were pointed and people were accused of "not working on a branch" even though they did at the time. I strongly encourage the ImgLib2 admins to use "Create a merge commit." This can still be combined with a rebase first, if you want the topic branch to sit on top the latest master. The goal is to _avoid fast forward merges_.

hanslovsky:
@ctrueden sorry for the confusion. I was advised to rebase my branch, then the PR would be merged with a commit message. That was before github had options to rebase through the web interface.

axtimwalde:
@ctrueden @hanslovsky the history of this pull request is painfully long and of limited educational value.  It is certainly possible to ask @gselzer to rebase this post-fact into meaningful dissection supporting units but I also think that we all have a lot of stuff to do and this is not very helpful.  I would therefore vote for squash and merge.

ctrueden:
@axtimwalde I don't see why further rebasing would be needed. The commits are already small and clean and granular. It sounds like your beef is really just that _there are a lot of them_ but I think that _helps_ bisectability, rather than harming it. There are no harmful intermediate states, are there? Why not just merge it?

hanslovsky:
@axtimwalde @tpietzsch Should I go ahead and squash and merge?

axtimwalde:
@hanslovsky yes.

hanslovsky:
Thanks @gselzer

PR #33, opened 2017-03-06
Add single-threaded version to LocalExtrema and allow for caller specified neighborhoods.

hanslovsky:
This commit adds:
  - shape parameter for callers to specify local neighborhood
  - Single-threaded version of findLocalExtrema
  - Parameter for caller to specify number of tasks.
  - Note in the JavaDoc to inform user that border pixels are ignored.
  - Relax the bounds on generic parameter T: Was Comparable< T >, now unbounded.

The original interface is preserved for a non-breaking change.

See also:
https://gitter.im/imglib/imglib2?at=58bd72101465c46a56f2ae9c

### DIFF
--- a/src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
+++ b/src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
@@ -138,7 +138,7 @@ public class LocalExtrema
 		final int nDim = source.numDimensions();
 		// Get biggest dimension after border subtraction. Parallelize along
 		// this dimension.
-		final int maxSizeDim = IntStream.range( 0, nDim ).mapToObj( d -> new ValuePair<>( d, source.dimension( d ) - 2 * borderSize[ d ] ) ).min( ( p1, p2 ) -> Long.compare( p1.getB(), p2.getB() ) ).get().getA();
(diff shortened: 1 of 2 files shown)


src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
```diff
@@ -182,32 +183,72 @@
...
+	 *            <code>shape</code> does not skip the center pixel.
+	 * @param shape
+	 *            Defines the local neighborhood.
+	 * @param service
+	 *            {@link ExecutorService} handles parallel tasks
+	 * @param numTasks
+	 *            Number of tasks for parallel execution
+	 * @return {@link ArrayList} of extrema
+	 */
+	public static < P, T > ArrayList< P > findLocalExtrema(
```
tinevez: Why return an `ArrayList` and not a `List` in the public API?

hanslovsky: I personally prefer to be as generic as possible with the input parameters and as specific as possible with the return values. The caller can still write `List< P > extrema = findLocalExtrema(...)`. Returning `List` would, of course, allow to change the implementation later on. This is, however, not an interface but a (set of) static method and the implementation details are unlikely to change.

hanslovsky:
Added more comprehensive tests and code coverage including the previously existing interface:

Code coverage is now at 88% for LocalExtrema.java, only missing:
 - Exception handlings for parallel execution
 - assert statements
 - MinimumCheck class

PR #35, opened 2017-03-20
Add Distance transform of sampled functions.

hanslovsky:
This PR adds a distance transform of sampled functions as described in:
https://cs.brown.edu/~pff/dt/
The input images can have arbitray scalar values and need not be binary, e.g. edge probability maps instead of binary edge maps.

 - All function overloads can be called in parallel (passing an `ExecutorService` and the number of tasks) or sequentially. The tasks will be parallelized along the longest dimension of the input.
 - The type of distane (e.g. L1 or Euclidian) is specified by either `enum DISTANCE_TYPE` or an implementation of the `Distance` interface.
 - When the distance type is specified by `DISTANCE_TYPE` the user can specify (anisotropic) weights.
 - The distance transform can be written into the source (only `source` as input), or the user can pass `tmp` and/or `target` input parameters. Pass both `tmp` and `target` is useful when the intermediate results should have higher precision than the final result in `target`.


hanslovsky:
I made the tests more comprehensive (fixed a bug on the way) and checked code coverage. It is at `93.2%` for the `net.imglib2.algorithm.morphology.distance` package and I miss only:
 - One constructor of `EuclidianDistanceAnisotropic`
 - 5 `DistanceTransform.transform` methods that take the `Distance` interface as a paremeter but they all delegate to tested code.
 - Certain branches in the sanity checks for inputs and default cases for switch statements that have no effect.

With that coverage I am confident that this is good code and I'll do a rebase + new pull request that is ready to merge.

axtimwalde:
@hanslovsky  Excellent!

PR #36, opened 2017-03-21
Add Distance transform of sampled functions

hanslovsky:
This PR adds a distance transform of sampled functions as described in:
https://cs.brown.edu/~pff/dt/
The input images can have arbitray scalar values and need not be binary, e.g. edge probability maps instead of binary edge maps.

 - All function overloads can be called in parallel (passing an `ExecutorService` and the number of tasks) or sequentially. The tasks will be parallelized along the longest dimension of the input.
 - The type of distane (e.g. L1 or Euclidian) is specified by either `enum DISTANCE_TYPE` or an implementation of the `Distance` interface.
 - When the distance type is specified by `DISTANCE_TYPE` the user can specify (anisotropic) weights.
 - The distance transform can be written into the source (only `source` as input), or the user can pass `tmp` and/or `target` input parameters. Pass both `tmp` and `target` is useful when the intermediate results should have higher precision than the final result in `target`.

This commit replaces PR #35

### DIFF
--- a/src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
+++ b/src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
@@ -1,0 +1,804 @@
+/*
+ * #%L
+ * ImgLib2: a general-purpose, multidimensional image processing library.
+ * %%
+ * Copyright (C) 2009 - 2016 Tobias Pietzsch, Stephan Preibisch, Stephan Saalfeld,
+ * John Bogovic, Albert Cardona, Barry DeZonia, Christian Dietz, Jan Funke,
+ * Aivar Grislis, Jonathan Hale, Grant Harris, Stefan Helfrich, Mark Hiner,
+ * Martin Horn, Steffen Jaensch, Lee Kamentsky, Larry Lindsey, Melissa Linkert,
+ * Mark Longair, Brian Northan, Nick Perry, Curtis Rueden, Johannes Schindelin,
+ * Jean-Yves Tinevez and Michael Zinsmaier.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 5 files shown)


src/test/java/net/imglib2/algorithm/morphology/distance/DistanceTransformTest.java
```diff
@@ -0,0 +1,305 @@
...
+	private static < T extends RealType< T >, U extends RealType< U > > void distanceTransform(
+			final Img< T > source,
+			final Img< U > target,
+			final DistanceFunctor d
+			)
+	{
+		for ( final Cursor< U > targetCursor = target.cursor(); targetCursor.hasNext(); )
+		{
+			final U t = targetCursor.next();
+			double tmp = Double.MAX_VALUE;
```
tpietzsch: use Double.POSITIVE_INFINITY instead?

src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
```diff
@@ -0,0 +1,804 @@
...
+			final RealComposite< DoubleType > envelopeIntersectLocation,
+			final Distance d,
+			final int dim,
+			final long size )
+	{
+		long k = 0;
+
+		lowerBoundDistanceIndex.get( 0 ).set( 0 );
+		envelopeIntersectLocation.get( 0 ).set( -1e20 );
+		envelopeIntersectLocation.get( 1 ).set( +1e20 );
```
tpietzsch: Is there any reason not to use `Double.NEGATIVE_INFINITY` and `Double.POSITIVE_INFINITY` here? (I tried and at least the test passes.)

src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
```diff
@@ -0,0 +1,804 @@
...
+			for ( double envelopeValueAtK = envelopeIntersectLocation.get( k ).get(); s <= envelopeValueAtK; envelopeValueAtK = envelopeIntersectLocation.get( k ).get() )
+			{
+				--k;
+				envelopeIndexAtK = lowerBoundDistanceIndex.get( k ).get();
+				s = d.intersect( envelopeIndexAtK, source.get( envelopeIndexAtK ).getRealDouble(), position, sourceAtPosition, dim );
+			}
+			++k;
+			lowerBoundDistanceIndex.get( k ).set( position );
+			envelopeIntersectLocation.get( k ).set( s );
+			envelopeIntersectLocation.get( k + 1 ).set( 1e20 );
```
tpietzsch: same as above: `Double.POSITIVE_INFINITY`?

tpietzsch:
@hanslovsky This looks fantastic!

Is there any reason you use `1e20` instead of `Double.POSITIVE_INFINITY`? (I also made the comment in the code.)

src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
hanslovsky: I will add that suggestion.

src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
hanslovsky: Same as above.

hanslovsky:
Thank you @tpietzsch !

tpietzsch:
@hanslovsky thank you!

PR #37, opened 2017-03-21
Add single-threaded version to LocalExtrema and allow for caller specified neighborhoods.

hanslovsky:
This commit adds:
  - shape parameter for callers to specify local neighborhood
  - Single-threaded version of findLocalExtrema
  - Parameter for caller to specify number of tasks.
  - Note in the JavaDoc to inform user that border pixels are ignored.
  - Relax the bounds on generic parameter T: Was Comparable< T >, now unbounded.

The original interface is preserved for a non-breaking change.

Code coverage is now at 88% for LocalExtrema.java, only missing:
 - Exception handlings for parallel execution
 - assert statements
 - MinimumCheck class

See also:
https://gitter.im/imglib/imglib2?at=58bd72101465c46a56f2ae9c

This PR is a rebase of and closes #33


src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
```diff
@@ -92,79 +100,370 @@
...
+				return findLocalExtrema( source, new FinalInterval( localMin, localMax ), localNeighborhoodCheck, shape );
+			} );
 		}
-		for ( final Future< Void > f : futures )
+
+		final ArrayList< P > extrema = new ArrayList<>();
+		// TODO It is probably better to throw exception than to use try/catch
+		// block and return potentially incomplete/inconsistent list of extrema.
+		// Returning an empty list on exception could be a compromise without
+		// changing the interface.
```
tpietzsch: I agree, throwing an exception is the better way!
Let's change the API, and add a deprecated version with the old signature that prints the exception and returns null.

src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
```diff
@@ -92,79 +100,370 @@
...
+	 * @param nDim
+	 *            Number of dimensions.
+	 * @return The required border size for the local neighborhood specified by
+	 *         <code>shape</code>
+	 */
+	public static long[] getRequiredBorderSize( final Shape shape, final int nDim )
+	{
+		final RandomAccessible< Neighborhood< Object > > neighborhood = shape.neighborhoodsRandomAccessible( ConstantUtils.constantRandomAccessible( new Object(), nDim ) );
+		final long[] min = LongStream.generate( () -> Long.MAX_VALUE ).limit( nDim ).toArray();
+		final long[] max = LongStream.generate( () -> Long.MIN_VALUE ).limit( nDim ).toArray();
```
tpietzsch: I probably sound like a grumpy old man here, but what's wrong with `Arrays.fill(max, Long.MIN_VALUE)`?
:-)

src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
```diff
@@ -92,79 +100,370 @@
...
+		final Interval bb = neighborhood.randomAccess().get().getStructuringElementBoundingBox();
+
+		for ( int d = 0; d < nDim; ++d ) {
+			min[ d ] = Math.min( bb.min( d ), min[ d ] );
+			max[ d ] = Math.max( bb.max( d ), max[ d ] );
+		}
+
+		final long[] borderSize = IntStream.range( 0, nDim ).mapToLong( d -> Math.max( max[ d ], -min[ d ] ) ).toArray();
```
tpietzsch: Similar here. Why not simply put `borderSize[ d ] = Math.max( max[ d ], -min[ d ] );` into the above for loop? To me it looks more readable, but that is of course a personal preference...

(Also it's probably much faster, but that isn't important here)

src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
```diff
@@ -92,79 +100,370 @@
...
-
-		final Interval full = Intervals.expand( img, -1 );
-		final int n = img.numDimensions();
-		final int splitd = n - 1;
-		// FIXME is there a better way to determine number of threads
+		final RectangleShape shape = new RectangleShape( 1, true );
+		final long[] borderSize = getRequiredBorderSize( shape, source.numDimensions() );
+		final int nDim = source.numDimensions();
+		// Get biggest dimension after border subtraction. Parallelize along
+		// this dimension.
```
tpietzsch: I'm not sure this is always a good choice. Assuming some kind of flattened array, where the highest dimension is the slowest changing, it is probably preferable to choose that one because of cache utilisation (even if the extent is larger along another dimension). It would be interesting to see a benchmark.

In the end, of course it can be a rotated view or the highest dimension may have extent 1, etc...

Can you add a version where the parallelisation dimension is a parameter?

src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
hanslovsky: Sounds reasonable to me! I will add that. 

As we are at interface changes already, what about @tinevez suggestion in #33 :
https://github.com/imglib/imglib2-algorithm/pull/33#pullrequestreview-25614889
We could return a `List` in the new interface but cast to `ArrayList` before returning in the deprecated version because we know the implementation details.
I still think that we should stick with `ArrayList` because that is unlikely to change but I'd like to hear your opinion on it.

src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
hanslovsky: Because Java still lacks a function that generates an array that's filled with a constant value in a single line, something like `Arrays.create( int length, long value ) would be great. This is my little workaround for that. ;) At least for me, readability is improved wthat way.

src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
hanslovsky: Again, I prefer the one-liner but that might be because I am coming from the python/numpy world originally where I would do something like `numpy.maximum( max, -min )`

src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
hanslovsky: Good point, I will add that and this method should delegate to it!

src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
hanslovsky: In the original implementation the task was split along the last dimension. Should I revert to that for the deprecated version for consistency?

tpietzsch: And yes, I would revert to splitting along the last dimension

hanslovsky: I will change it to `List`, then. And revert to spliiting along the last dimension for the deprecated version.

hanslovsky:
Can this be merged now?

src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
```diff
@@ -92,79 +100,457 @@
...
+			final RandomAccessibleInterval< T > source,
+			final LocalNeighborhoodCheck< P, T > localNeighborhoodCheck,
+			final Shape shape )
+	{
+
+		final long[] borderSize = getRequiredBorderSize( shape, source.numDimensions() );
+
+		assert Arrays.stream( borderSize ).min().getAsLong() >= 0: "Border size cannot be smaller than zero.";
+
+		// TODO use Intervals.expand once it is available in non-SNAPSHOT
```
tpietzsch: I just released imglib2-4.6.0. The `Intervals.expand` should be now available.

tpietzsch:
I just released imglib2-4.6.0. The `Intervals.expand` should be now available. Could you address that one TODO item referring to that?

Then it's ready to go!

(There is a javadoc link error in the class comment. It was there already before, but maybe while you're at it, could you fix it?)

hanslovsky:
Great, I'll take care of the `Intervals.expand` and the javadoc link error.

axtimwalde:
I think that overriding the one dependency version that you care for is ok?

hanslovsky:
If that is the general policy, I can of course go ahead and add the required changes. Is that ok only while on a SNAPSHOT version, i.e. overriding the dependency version should not happen in a release version?

ctrueden:
@hanslovsky Not sure I understand your question correctly, but for SciJava projects in `scijava`, `imagej`, `scifio` and `fiji`, the policy is that all commits merging to master must reproducibly compile with passing tests—otherwise it is a "bug". So that means you cannot have a SNAPSHOT dependency in any such commits, even if a later commit "fixes" it. Any commit adding a SNAPSHOT dependency should be labeled as WIP and fixed (i.e. rebased) to not be a WIP before that branch can be merged.

Does that answer your question?

hanslovsky:
Thanks for the clarification @ctrueden 
The reason why I asked was exactly that version skew that you were referring to in your post but that does not seem to be a big concern here, so I will go ahead and update my branch with the suggested changes @tpietzsch @axtimwalde

ctrueden:
> version skew that you were referring to in your post but that does not seem to be a big concern here

Well, it is not a concern *here*. The problems typically happen downstream somewhere. For example, suppose I have a project which extends the latest `pom-scijava`, but wants to use this new version of `LocalExtrema` you've added here. So I pin to the latest release of `imglib2-algorithm`, leaving the rest of the versions as default inherited from `pom-scijava`. Now my code will fail, because it will use a too-old version of `imglib2`. This will be surprising to many people, and in some cases very subtle and dangerous when, instead of an explicit API breakage/mismatch, we instead inherit some old version of a component with a subtle but devastating bug that is not easily noticed by a human.

hanslovsky:
I saw that `pom-imglib2-10.0.1` was released that has `pom-scijava-19.0.0` as parent, which defines `<imglib2.version>4.6.0</imglib2.version>`. Can we update the parent version of `imglib2-algorithm` to `10.0.1` so we can merge this without overriding `imglib2.version` and avoiding potential version skew?

hanslovsky:
This happened in #50 I merged master into this branch so this should be ready to be merged now without version skew.

hanslovsky:
@imglib/admins can we merge this? From the discussion it seems that the imglib2 dependency version was the only remaining issue, which seems to be resolved now. I do not want to merge my own PR (as a non-owner), though.

axtimwalde:
Pleas merge.

PR #38, opened 2017-03-21
Remove duplicate <groupId> tag.

hanslovsky:
Duplicate caused a warning in eclipse. <groupId> already specified in parent.


tpietzsch:
thanks @hanslovsky

PR #39, opened 2017-04-06
Bresenham circle.

tinevez:
Each point of the circle is iterated exactly once, and there is no "hole" in the circle. Contrary to the algorithm in Wikipedia, the circles generated by this cursor are "slim": each pixel of the circle is connected with 8-connectivity. For instance, a bitmap excerpt from such a circle looks like this:
```
 ..........
 OOO.......
 ...OO.....
 .....O....
 .....O....
```


imagejan:
@tinevez is there a reason why you were closing this?

tinevez:
Yes! I want to make it better.

ctrueden:
Just in case you didn't know: you can force push your changes onto this same branch, and then reopen the PR, instead of opening a new one.

PR #44, opened 2017-06-22
Differences

tibuch:
* Remove the nice but slow implementation of central difference.
* Add forward difference
* Add backward difference

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>net.imglib2</groupId>
 		<artifactId>pom-imglib2</artifactId>
-		<version>10.0.1</version>
+		<version>10.0.2</version>
 		<relativePath />
 	</parent>
 
--- a/src/main/java/net/imglib2/algorithm/gradient/PartialDerivative.java
+++ b/src/main/java/net/imglib2/algorithm/gradient/PartialDerivative.java
@@ -54,6 +54,8 @@ import net.imglib2.view.Views;
 /**
  * @author Tobias Pietzsch
  * @author Philipp Hanslovsky
+ * @author Tim-Oliver Buchholz
+ * @author Matthias Arzt
  *
  */
 public class PartialDerivative
--- a/src/main/java/net/imglib2/algorithm/gradient/PartialDerivative.java
+++ b/src/main/java/net/imglib2/algorithm/gradient/PartialDerivative.java
@@ -60,7 +60,10 @@ public class PartialDerivative
 {
 	// nice version...
 	/**
-	 * Compute the partial derivative of source in a particular dimension.
+	 * Compute the partial derivative (central difference approximation) of source
+	 * in a particular dimension:
+	 * {@code d_f( x ) = ( f( x + e ) - f( x - e ) ) / 2},
+	 * where {@code e} is the unit vector along that dimension.
 	 *
 	 * @param source
 	 *            source image, has to provide valid data in the interval of the
@@ -85,7 +88,10 @@ public class PartialDerivative
 
 	// parallel version...
 	/**
-	 * Compute the partial derivative of source in a particular dimension.
+	 * Compute the partial derivative (central difference approximation) of source
+	 * in a particular dimension:
+	 * {@code d_f( x ) = ( f( x + e ) - f( x - e ) ) / 2},
+	 * where {@code e} is the unit vector along that dimension.
 	 *
 	 * @param source
 	 *            source image, has to provide valid data in the interval of the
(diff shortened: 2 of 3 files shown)


tpietzsch:
I would rather leave the nice and slow version for educational purposes. Just comment it out, but leave it in.

tibuch:
Removed the remove commit.

hanslovsky:
Please add test cases for small 2D and 3D images.

src/main/java/net/imglib2/algorithm/gradient/PartialDerivative.java
```diff
@@ -234,4 +234,142 @@
 			}
 		}
 	}
+	
+	/**
+	 * Compute the backward difference of source in a particular dimension.
```
hanslovsky: Please at some JavaDoc math that specifies exactly what is being computed, like
https://en.wikipedia.org/wiki/Finite_difference#Forward,_backward,_and_central_differences

src/main/java/net/imglib2/algorithm/gradient/PartialDerivative.java
```diff
@@ -234,4 +234,142 @@
...
+			else {
+				result.fwd(0);
+				back.fwd(0);
+				current.fwd(0);
+			}
+		}
+	}
+	
+	/**
+	 * Compute the forward difference of source in a particular dimension.
```
hanslovsky: Please at some JavaDoc math that specifies exactly what is being computed, like
https://en.wikipedia.org/wiki/Finite_difference#Forward,_backward,_and_central_differences

src/main/java/net/imglib2/algorithm/gradient/PartialDerivative.java
tibuch: Using the LoopBuilder implementation by @maarzt makes the code much more readable. If necessary I can add more java doc.

hanslovsky: Having the math in the doc would be very helpful, either in code
```
d_f( x ) = f( x + h ) - f( x )
```
or tex
[](https://en.wikipedia.org/wiki/Finite_difference#Forward,_backward,_and_central_differences)
(examples for forward difference).

src/main/java/net/imglib2/algorithm/gradient/PartialDerivative.java
hanslovsky: Having the math in the doc would be very helpful, either in code
```
d_f( x ) = f( x + h ) - f( x )
```
or tex
[](https://en.wikipedia.org/wiki/Finite_difference#Forward,_backward,_and_central_differences)
(examples for forward difference).

pom.xml
```diff
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>net.imglib2</groupId>
 		<artifactId>pom-imglib2</artifactId>
-		<version>10.0.1</version>
+		<version>10.0.2</version>
```
hanslovsky: Why do we need the bump here? Is it for the `LoopBuilder`?

src/test/java/net/imglib2/algorithm/gradient/PartialDerivativeTest.java
```diff
@@ -0,0 +1,94 @@
+package net.imglib2.algorithm.gradient;
```
hanslovsky: Please run imglib2 eclipse [formatter](https://github.com/imglib/imglib2/blob/master/doc/imglib2-eclipse-mars-clean-up-formatter-style.xml) on this file.

hanslovsky:
Added some minor comments but then it looks good to merge for me.

src/main/java/net/imglib2/algorithm/gradient/PartialDerivative.java
maarzt: The function `f` in the equation `d_f( x ) = f( x + h ) - f( x )` has an one dimensional domain. If we want to describe what `gradientForwardDifference` actually computes on 2d, 3d or Nd images, we would need a more complicated equation.
I think the resulting math will be rather confusing. @hanslovsky Please proof me wrong!

pom.xml
maarzt: Yes, it's because this makes `LoopBuilder` available.
I changed commit message, to include that information.

src/test/java/net/imglib2/algorithm/gradient/PartialDerivativeTest.java
maarzt: done

hanslovsky:
Looks good to me now. Will leave open for a bit for more comments/concerns and then merge eventually.

PR #50, opened 2018-01-17
use the current version of pom-imglib2 as parent. Currently, it was b…

tpietzsch:
…uilt against imglib2-3.3.0, which failed when any algorithm instantiates a CellImg itself due to a change in the constructor methods.

This is critical!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>net.imglib2</groupId>
 		<artifactId>pom-imglib2</artifactId>
-		<version>9.0.0</version>
+		<version>10.0.1</version>
 		<relativePath />
 	</parent>
 


tpietzsch:
will release later today

StephanPreibisch:
Hi @tpietzsch, I made the same commits for imglib2-algorithm-gpl and imglib2-algorithm-fft ... would be great to release them all together

StephanPreibisch:
@tpietzsch and imglib2-ij

tpietzsch:
all released

StephanPreibisch:
Awesome, thanks @tpietzsch

StephanPreibisch:
I would upload them to Fiji to fix the behavior ... any objections @tpietzsch @ctrueden @axtimwalde

imagejan:
Hm, before they're uploaded to the *Java-8* update site, I'd like to see the versions updated in `pom-scijava` first, so that the user version is not "newer" than what you get when using the latest `pom-scijava` parent.

What do you think?

https://github.com/scijava/pom-scijava/blob/9aaf167db313703e6db9297b6bdc121406181a94/pom.xml#L301-313

StephanPreibisch:
Well, right now things are crashing ... @ctrueden, what do you think, how should we proceed?

imagejan:
@StephanPreibisch wrote:
> Well, right now things are crashing ... @ctrueden, what do you think, how should we proceed?

To help sticking to ImageJ [development principles](http://imagej.net/Development_Lifecycle#Phase_4:_Managed) ...

>To ensure users and developers see the same functionality, components should only be uploaded to the core update sites when their version is also updated in the pom-scijava BOM.

... I created PR https://github.com/scijava/pom-scijava/pull/56 to update the component versions in `pom-scijava`. As soon as this is merged, it's good to upload, I'd say 😃.

ctrueden:
I merged @imagejan's PR, and am releasing pom-scijava version 19.1.1 now. Note that the previous release of pom-scijava, 19.1.0, failed due to a "504 Gateway Timed Out" error. I am hoping that was a transient error with Sonatype's server, and does not happen again this time. 🤞

ctrueden:
The release of pom-scijava 19.1.1 was successful. This parent POM can now be used to inherit the (hopefully!) working versions of the imglib2 components.

PR #53, opened 2018-01-22
Add utility methods to parallelize over blocks

hanslovsky:
This commits adds 
 - `ParallelizeOverBlocks.parallelize` - parallelize a task over blocks and return futures
 - `ParallelizeOverBlocks.parallelizeAndWait` - parallelize a task over blocks and wait for the result
with various signatures. The three signatures are distingished by the way the blocks for parallelization are specified:

 1. `Interval interval`, `int[] blockSize`
 2. `long[] min`, `long[] max`, `int[] blockSize`
 3. `List< Interval > blocks`

This is a common pattern that I have used multiple times and I think that imglib2-algorithm is the adequate package for such methods.

`BlockOffsets.java` contains helper methods for generating lists of blocks from intervals and block sizes.


hanslovsky:
@imglib/admins please comment

axtimwalde:
Sounds reasonable but in practice I always needed more.  We should have a look at this class
https://github.com/saalfeldlab/hot-knife/blob/master/src/main/java/org/janelia/saalfeldlab/hotknife/util/Grid.java
and add what's missing (i.e. cropping the grid intervals to input interval boundaries and related things).

hanslovsky:
- Rename `BlockOffsets` to `Grids` to be closer to imglib2 namespace (`CellGrid` etc)
 - Add tests (detected a bug for min != 0)
 - new convenience methods for:
   - retrieving complete intervals (instead of just min)
   - retrieving complete intervals and position within cell grid
Unfortunately, it is not possible to use `CellGrid` here because we have to consider a non-zero min, unless we have specific implementations of `CreateAndCropBlockToFitInterval` and `GetGridCoordinates` for zero min and non-zero min, respectively. This does not sound reasonable to me, though.

We should still consider moving the `Cells` class into imglib2 core, though, as discussed in person.

hanslovsky:
I had a more common use case like this:
```java
void forEachOffset( long[] min, long[] max, int[] step, Consumer< long[] > doAtOffset );
```
This can be generalized to allow for use with `RandomAcess`, for example. I will add this before the PR gets merged.

hanslovsky:
Any comments/objections against a merge? I would like to go ahead and merge this.

hanslovsky:
@imglib/admins if there are no further objections or comments, I will go ahead and merge this PR.

PR #57, opened 2018-03-20
Remove imagej-common dependency

### LINKED ISSUE #43: Remove imagej-common dependency
We should get rid of the imagej-common dependency. It is only needed in DoG to get image calibration from `net.imagej.space.LinearSpace`. This should not happen here. Just move it to ops.

hanslovsky:
Fixes #43

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -219,12 +219,6 @@ Jean-Yves Tinevez and Michael Zinsmaier.</license.copyrightOwners>
 			<artifactId>imglib2-realtransform</artifactId>
 		</dependency>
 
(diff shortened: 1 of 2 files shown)


imagejan:
According to SemVer, removing public API would require (at least) a version bump to `0.9.0-SNAPSHOT`, no?

hanslovsky:
Currently, we are still at MAJOR version 0, which 
 > [is for initial development. Anything may change at any time. The public API should not be considered stable.](https://semver.org/#spec-item-4)

PR #58, opened 2018-03-22
 Update FloodFill interface

### LINKED ISSUE #55: FloodFill interface is not intuitive
Two of the parameters of [`FloodFill.fill`](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/fill/FloodFill.java#L173) are not intuitive:
  - [`Filter< Pair< T, U >, Pair< T, U > > filter`](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/fill/Filter.java): Compares a pair of image and mask pixels with a reference pair to determine whether or not the pixel should be added to the floodfill. A `BiPredicate< T, U >` that only checks the current pixel is more intuitive and the caller will always know the reference values anyway, and can include them in the `BiPredicate` if necessary.
  - [`Writer< U > writer`](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/fill/Writer.java): Interface that defines how to update a target value with a source value. Again, the caller has access to the source value and it would thus be more intuitive to use a `Consumer< U >` in this case.

I suggest we deprecate the `Filter` and `Writer` interfaces, and all method signatures that use these interfaces. We add new signatures that take `BiPredicate` and `Consumer` instead, and convenience methods as appiclable.

hanslovsky:
Fixes #55 

Current interface is not intuitive:
 - Deprecate confusing interfaces:
   - Filter
   - Writer
   - TypeWriter
 - Replace by:
   - BiPredicate
   - Consumer
   - appropriate lambda
 - Deprecate current interfaces and add new interfaces using BiPredicate/Consumer

### DIFF
--- a/src/main/java/net/imglib2/algorithm/fill/Filter.java
+++ b/src/main/java/net/imglib2/algorithm/fill/Filter.java
@@ -35,8 +35,8 @@
 package net.imglib2.algorithm.fill;
 
 /**
- * Interface for comparing {@link T} t and {@link U} u and accepting
- * them as equivalent in a sense specified by implementation thereof.
+ * Interface for comparing {@link T} t and {@link U} u and accepting them as
+ * equivalent in a sense specified by implementation thereof.
  *
  * @author Philipp Hanslovsky
  * @author Stephan Saalfeld
@@ -44,6 +44,7 @@ package net.imglib2.algorithm.fill;
  * @param <T>
  * @param <U>
  */
+@Deprecated
 public interface Filter< T, U >
 {
 	boolean accept( T t, U u );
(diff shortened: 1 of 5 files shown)


tpietzsch:
I had a very brief look and it looks good to me. Fine to merge it.

Independent of this fix, in the javadoc of generic parameters, I would say what they represent, not the restrictions.
https://github.com/imglib/imglib2-algorithm/pull/58/files#diff-97cf72110dceb6de78c7e1473f2d8162R90
```java
 * @param <T>
 *            T implements {@code Type<U>}.
```
just repeats (in this case wrong) what the method signature says anyways.
Something like
```java
 * @param <T>
 *            input pixel type.
 * @param <U>
 *            fill label type.
```
would be more helpful.

hanslovsky:
Will update the JavaDoc and merge.

PR #60, opened 2018-04-13
Constrained Ransac based ellipse fitting, sphere fitting on 4 points

kapoorlab:
Constrain Ransac to find ellipses on candidate points and determine their points of intersections, construct tangents at those points and determine the angle of intersection between the ellipses. 

Fit 3D ellipsoids satisfying the same constraints.

Given 3/4 points, fit a Circle/sphere using yury petrov regression for local curvature measurements.


tpietzsch:
This PR in `src/main/java/subpixelLocalization/GaussianLineds.java` contains copy-pasted GPL code. imglib-algorithm is BSD licensed, so this is not possible.

kapoorlab:
I can remove the subpixel localization folder out of the repo, would that help?

kapoorlab:
I will make the changes to resolve the conflicts and then make a new pull request.

imagejan:
@kapoorlab you can always force-push to your own topic branch to update the pull request. There should be no need to open a new one.

PR #65, opened 2018-05-22
Fast gauss

maarzt:
Add convolution package that's more general than gauss3. (Support convolvers not based on symmetric kernels)

& add fast gauss by vlado


maarzt:
It's still work in progress. I accidentally created this PR.

PR #71, opened 2018-08-02
Improve distance transform usability

### LINKED ISSUE #69: Improve DistanceTransform usability
@tischi recently contacted me about using [`DistanceTransform`](https://github.com/imglib/imglib2-algorithm/blob/47cd6ed5c97cca4b316c92d4d3260086a335544d/src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java) and had a few questions about it. After helping him out we compiled a short list of improvements that would make calling `DistanceTransform` easier:
 - [x] Documentation should clearly state that Euclidian distance is actually squared Euclidian distance.
 - [x] There should be convenience methods for binary distance transform (converting `true` to zero and `false` to positive infinity)
 - [x] Information about distance transform from **sampled functions** should be added to each individual overload (not only in class doc) so callers can see it when they auto-complete in eclipse
 - [x] add comment that weights in squared Euclidian transform are not being squared (while coordinates are), i.e. callers need to pass squared weights, e.g. to adjust for pixel anisotropy

hanslovsky:
Addresses requested usability improvements of `DistanceTransform` as requested in #69 
Fixes #69

### DIFF
--- a/src/main/java/net/imglib2/algorithm/morphology/distance/Distance.java
+++ b/src/main/java/net/imglib2/algorithm/morphology/distance/Distance.java
@@ -45,7 +45,7 @@ package net.imglib2.algorithm.morphology.distance;
  * {@code d' = f(x - x0) + y0}, must have exactly one intersection point (for
  * each dimension):
  * </p>
- * 
+ *
  * <pre>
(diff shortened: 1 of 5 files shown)


hanslovsky:
Merging after consulting with @axtimwalde

PR #75, opened 2018-11-15
ImgMath fixes

acardona:
Two fixes for ImgMath:
 1. A critical issue, that appears only in scripting languages that fail to find the 2-argument constructor of many functions, and instead use the varargs argument with 2 arguments. The Util.wrapMap was not ready to work with less than 3 arguments.
2. A minor issue: lack of two static functions in ImgMath: img (to wrap a RandomAccessibleInterval) and number (to wrap a number). These two enable convenient operations like e.g. copying and image into another, and filling an image with a number.

### DIFF
--- a/src/main/java/net/imglib2/algorithm/math/Add.java
+++ b/src/main/java/net/imglib2/algorithm/math/Add.java
@@ -26,8 +26,10 @@ public final class Add extends ABinaryFunction
 			final O tmp,
 			final Map< String, O > bindings,
 			final Converter< RealType< ? >, O > converter,
-			Map< Variable< O >, OFunction< O > > imgSources )
+			final Map< Variable< O >, OFunction< O > > imgSources )
 	{
-		return new Addition< O >( tmp.copy(), this.a.reInit( tmp, bindings, converter, imgSources ), this.b.reInit( tmp, bindings, converter, imgSources ) );
+		return new Addition< O >( tmp.copy(),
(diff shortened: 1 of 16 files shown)


acardona:
Thanks for looking at the code changes.

The ```VarargsFunction.wrapMap()``` has to take an ```Object[]```, because it is invoked by ```ABinaryFunction``` which itself takes a varargs, which within the code block appears as an ```Object[]```. Otherwise, it would be passing an Object[][]. So all is fine as is, and works correctly, which is the most important.

tpietzsch:
> The `VarargsFunction.wrapMap()` has to take an `Object[]`, because it is invoked by `ABinaryFunction` which itself takes a varargs, which within the code block appears as an `Object[]`. Otherwise, it would be passing an Object[][]. So all is fine as is, and works correctly, which is the most important.

Hmm, actually, did you try that???

```java
	private static void ao( final Object... i ) {
		System.out.println( Arrays.toString( i ) );
	}

	private static void bo( final Object... i ) {
		ao( i );
	}

	public static void main( final String[] args ) {
		ao( "x", 1, 2 );
		bo( "x", 1, 2 );
	}
```
prints
```
[x, 1, 2]
[x, 1, 2]
```
So it looks like `Object...` varargs will *not* be re-wrapped as `Object[][]`

acardona:
Good to know! I was for some reason convinced that's not the case.

I have some features pending for ImgMath, will test and address the varargs in the next PR.

PR #84, opened 2019-10-14
Fix Derivative of Gauss3 Kernel

maarzt:
I noticed that under certain circumstances the derivative of a image blurred with `Gauss3` might contain significant artifacts. These artifacts are caused by the truncation of the kernel, which results in a step at the end of the kernel. This "step" can be seen in a derivative of the image.

The problem is known and solved in ImageJ1 GaussianBlur: https://github.com/imagej/imagej1/blob/4a98636b35bf1f75a25062d2793e3558cdcc6291/ij/plugin/filter/GaussianBlur.java#L515-L518

This PR applies the same solution (as known from IJ1) to Gauss3. Tests to measure the quality of the Gaussian blur and it's first and second derivative are added too.


axtimwalde:
I think we should merge this? @tpietzsch @StephanPreibisch

tpietzsch:
👍

PR #86, opened 2020-06-25
bspline interpolation

bogovicj:
@axtimwalde @tpietzsch 

This pull request adds two kinds of BSpline interpolation. 

## 1) On the fly interpolation
The first is most similar to current interpolation methods:
```
int splineOrder;
boolean clipping;
int radius;
realImg = Views.interpolate( img, new BSplineInterpolatorFactory<>( splineOrder, clipping, radius ));
```
This computes the interpolation kernel on-the-fly at every `get`.  The kernel size is defined by `radius`.  `clipping=true` sets the interpolator to behave like the [ClampingNLinearInterpolator](https://github.com/bogovicj/imglib2/blob/b33186acc75abb8292da9865b6d100fc0787644d/src/main/java/net/imglib2/interpolation/randomaccess/ClampingNLinearInterpolatorRealType.java) 

## 2) Precomputed coefficients 
The second method precomputes BSpline coefficients and is substantially faster if `get` is called many times, but at the cost of memory - it allocates a `RandomAccessibleInterval<T>` the same size as the image to be interpolated.  `T` defaults to `DoubleType`.  Use it like this:

```
int splineOrder;
boolean clipping;
factory= new BSplineCoefficientsInterpolatorFactory<>( img, splineOrder, clipping );
realImg = Views.interpolate( img, factory);
```
here the factory constructor computes the coefficients - which is why it needs to be passed.  Unfortunately, this means that the `img` argument to `Views.interpolate` is meaningless.  For example, 
```
realImgFoo = Views.interpolate( img, factory);
realImgBar = Views.interpolate( someOtherImg, factory);
```
both result in the same thing.  We stuck with this to keep the api consistent.  No one should be doing the bottom thing anyway.

### Generating the coefficients
`BSplineDecomposition` and `BSplineCoefficientsInterpolator` give more control for those who need it.
For example
```
/*
 * Compute the coefficients over an arbitrary interval 
 */
BSplineDecomposition<T,S> decomp = new BSplineDecomposition<>( splineOrder,, extendedImg );
long[] min = Intervals.minAsLongArray( interval );
Img<S> coefficientsBase = coefficientFactory.create( interval );
coefficients = Views.translate( coefficientsBase, min );
decomp.accept( coefficients );

/*
 * Get a RealRandomAccess that uses the spline coefficients computed above
 * and using a custom extension of the coefficients
 */
RealRandomAccess interp = BSplineCoefficientsInterpolator.build( 
     splineOrder, 
     Views.extendMirrorSingle( coefficients), 
     new FloatType());
```


bogovicj:
Note: don't merge this quite yet, I'm going to add methodology to lazily compute the coefficients block-wise on demand.

tpietzsch:
> Note: don't merge this quite yet, I'm going to add methodology to lazily compute the coefficients block-wise on demand.

You can mark the PR as a draft while you work on it.
Then it cannot be merged yet.
It's on the right side in the Reviewers tab

Screenshot 2020-06-26 at 11 31 07

bogovicj:
@axtimwalde 
From the hotknife repo, I moved [`org.janelia.saalfeldlab.hotknife.util.Lazy`](https://github.com/saalfeldlab/hot-knife/blob/dc40394e576f40ae2d926c4d3f429bc64d494a6c/src/main/java/org/janelia/saalfeldlab/hotknife/util/Lazy.java) to `net.imglib2.lazy.Lazy`, removing some of the Ops- related methods

bogovicj:
TODO - make all added packages children of `net.imglib2.algorithm`

pom.xml
```diff
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.scijava</groupId>
 		<artifactId>pom-scijava</artifactId>
-		<version>27.0.1</version>
+		<version>28.0.0</version>
```
axtimwalde, line 8: Have you tried 29.0.0-beta-3?

src/main/java/net/imglib2/algorithm/lazy/Lazy.java
```diff
@@ -0,0 +1,148 @@
+/**
```
axtimwalde, line 1: Please change license to ImgLib2 license.  Permitted by the author ;).

src/test/java/net/imglib2/algorithm/bspline/GenBsplineKernels.java
```diff
@@ -0,0 +1,525 @@
...
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import net.imglib2.algorithm.interpolation.randomaccess.BSplineInterpolator;
+
+public class GenBsplineKernels {
+
+	public static void main(String[] args) throws IOException
```
axtimwalde, line 14: What is this?

src/test/java/net/imglib2/algorithm/bspline/GenBsplineKernels.java
```diff
@@ -0,0 +1,525 @@
...
+	}
+	
+	public static double kernelCenter( int n )
+	{
+		return (double)( n + 1 ) / 2.0;
+	}
+	
+	public static void tests() throws IOException
+	{
+//		System.out.println( fact( 5 ));
```
axtimwalde, line 173: Can this go?

src/test/java/net/imglib2/algorithm/bspline/GenBsplineKernels.java
```diff
@@ -0,0 +1,525 @@
...
+//
+		toCsv( 	"/home/john/tmp/bsplineKernelTests/ker5.csv", 
+				splinePieces( 5 ),
+				start, end, 0.1 );
+		
+	}
+	
+	public static void centeredPolysToCsv() throws IOException
+	{
+		double start = -7.0;
```
axtimwalde, line 236: And this?  Do you want to make this into a test?  That could be useful.

src/main/java/net/imglib2/algorithm/bspline/BSplineDecomposition.java
```diff
@@ -0,0 +1,582 @@
...
+	 * 
+	 * See Box 2 of Unser 1999
+	 */
+	protected static <T extends RealType<T>> double initializeCausalCoefficients(
+			final double z,
+			final double tolerance,
+			final int dimension,
+			final RandomAccess<T> dataAccess )
+	{
+		// TODO this method is wasteful if the interval is zero-extended
```
axtimwalde, line 496: Still a TODO?

src/main/java/net/imglib2/algorithm/bspline/BSplineDecomposition.java
```diff
@@ -0,0 +1,582 @@
...
+			S coef = destAccess.get();
+			coef.sub( previous );
+			coef.mul( -z );
+			previous.set( coef );
+
+			srcAccess.bck( dimension );
+			destAccess.bck( dimension );
+		}
+
+		// TODO may need to back things up to make sure we're in the right place
```
axtimwalde, line 427: Sounds scary.

src/test/java/net/imglib2/algorithm/bspline/GenBsplineKernels.java
bogovicj: old test. will remove

bogovicj: err, some of that is old tests. Other parts I want to keep - namely they generate the numbers that go [here](https://github.com/bogovicj/imglib2-algorithm/blob/bsplineInterpolation/src/main/java/net/imglib2/algorithm/bspline/BsplineKernel4.java#L18-L26), for example.  I'd want to keep that around somewhere

bogovicj: documented this class better.  Tests as junit now in their own class.

src/main/java/net/imglib2/algorithm/bspline/BSplineDecomposition.java
bogovicj: its not a TODO, but rather a comment.  So TODO removed.

src/main/java/net/imglib2/algorithm/bspline/BSplineDecomposition.java
bogovicj: obsolete comment removed.

pom.xml
imagejan: Can now be updated to 29.0.0

PR #89, opened 2020-11-04
mutual information and related methods

bogovicj:


src/main/java/net/imglib2/algorithm/stats/InformationMetrics.java
```diff
@@ -0,0 +1,185 @@
...
+		long ctot = 0;
+		for( int i = 0; i < ni; i++ )
+		{
+			count = subHistCount( hist, dim, i );
+			ctot += count;
+			double p = 1.0 * count / total;
+
+			if( p > 0 )
+				entropy -= p * Math.log( p );
+		}
```
imagejan, line 170: Is this something we can run multi-threaded using `LoopBuilder`? See also #83.

bogovicj: Good idea, I'll check :+1: thanks @imagejan

src/main/java/net/imglib2/algorithm/stats/InformationMetrics.java
```diff
@@ -0,0 +1,218 @@
...
+
+	/**
+	 * Returns the normalized mutual information of the inputs
+	 * @param rai the RandomAccessibleInterval
+	 * @param ra the RandomAccessible
+	 * @return the normalized mutual information
+	 */
+	public static <T extends RealType< T >> double normalizedMutualInformation(
+			final IterableInterval< T > dataA,
+			final IterableInterval< T > dataB,
```
hanslovsky, line 34: Is there a need to pass `IterableInterval`? Could it just be `Iterable`?

src/main/java/net/imglib2/algorithm/stats/InformationMetrics.java
```diff
@@ -0,0 +1,218 @@
...
+ * 
+ * @author John Bogovic
+ */
+public class InformationMetrics
+{
+
+	/**
+	 * Returns the normalized mutual information of the inputs
+	 * @param rai the RandomAccessibleInterval
+	 * @param ra the RandomAccessible
```
hanslovsky, line 29: Inconsistent with actual types of parameters in method signature

src/main/java/net/imglib2/algorithm/stats/InformationMetrics.java
```diff
@@ -0,0 +1,218 @@
...
+			final IterableInterval< T > dataA,
+			final IterableInterval< T > dataB,
+			final double histmin,
+			final double histmax,
+			final int numBins )
+	{
+		Real1dBinMapper< T > binMapper = new Real1dBinMapper< T >( histmin, histmax, numBins, false );
+		ArrayList< BinMapper1d< T > > binMappers = new ArrayList< BinMapper1d< T > >( 2 );
+		binMappers.add( binMapper );
+		binMappers.add( binMapper );
```
hanslovsky, line 108: Not tested but something like this should be possible:
```suggestion
		List< BinMapper1d< T > > binMappers = Arrays.asList( binMapper, binMapper );
```

src/main/java/net/imglib2/algorithm/stats/InformationMetrics.java
```diff
@@ -0,0 +1,218 @@
...
+
+		List< Iterable< T > > data = new ArrayList< Iterable< T > >( 2 );
+		data.add( dataA );
+		data.add( dataB );
```
hanslovsky, line 112: Not tested but something like this should be possible:
```suggestion
		List< Iterable< T > > data = Arrays.asList( dataA, dataB );
```

src/main/java/net/imglib2/algorithm/stats/InformationMetrics.java
```diff
@@ -0,0 +1,218 @@
...
+			IterableInterval< T > dataB,
+			double histmin, double histmax, int numBins )
+	{
+		HistogramNd< T > jointHist = jointHistogram( dataA, dataB, histmin, histmax, numBins );
+
+		double HA = marginalEntropy( jointHist, 0 );
+		double HB = marginalEntropy( jointHist, 1 );
+		double HAB = entropy( jointHist );
+
+		return HA + HB - HAB;
```
hanslovsky, line 72: This should call `mutualInformation( jointHist )` instead of duplicating the code.

PR #91, opened 2021-05-26
Image quality and segmentation metrics

jdeschamps:
A bunch of classes to compute image quality or segmentation metrics between two RAI. The results have been checked against original implementations (Cell Tracking Challenge, StarDist) or widely used librairies (scikit-image). 

**Image Quality:**

- _Mean squared error_ (MSE)
- _Peak signal-to-noise ratio_ (PSNR)
- _Structural similarity index measure_ (SSIM)

**Segmentation:**

- _SEG_ (from the [Cell Tracking Challenge](http://celltrackingchallenge.net/))
- LazySEG: _SEG_ calculation compatible with lazy loading and multithreading, accumulating the SEG score from each image pair added to the computation. 
- MultiMetrics: _TP, FP, FN, Precision, Recall, F1, Accuracy, mean matched and mean true IoU_  (inspired from [StarDist](https://github.com/stardist/stardist))
- LazyMultiMetrics: MultiMetrics calculations compatible with lazy loading and multithreading, accumulating the SEG score from each image pair added to the computation.


src/main/java/net/imglib2/algorithm/metrics/imagequality/MSE.java
```diff
@@ -0,0 +1,55 @@
+package net.imglib2.algorithm.metrics.imagequality;
```
imagejan, line 1: Can you add license headers by running `mvn license:update-file-header license:update-project-license`?

src/main/java/net/imglib2/algorithm/metrics/segmentation/SegmentationHelper.java
```diff
@@ -0,0 +1,68 @@
...
+	 * @param img
+	 * 		Image labeling from which to extract the occurring pixel values
+	 * @param <T>
+	 * 		Label type associated to the ground-truth
+	 * @param <I>
+	 * 		Image pixel type
+	 *
+	 * @return {@link Set} of occurring pixel values
+	 */
+	static < T, I extends IntegerType< I > > Set< I > getOccurringLabelSets( ImgLabeling< T, I > img )
```
imagejan, line 31: Isn't this something that [`LabelRegions#getExistingLabels()`](https://github.com/imglib/imglib2-roi/blob/f1e5daae75cab827790293b0bd9c9859f160d7c8/src/main/java/net/imglib2/roi/labeling/LabelRegions.java#L132-L136) already provides?

src/main/java/net/imglib2/algorithm/metrics/segmentation/SegmentationHelper.java
```diff
@@ -0,0 +1,68 @@
...
+	 * @param img
+	 * 		Image labeling
+	 * @param <T>
+	 * 		Label type associated to the ground-truth
+	 * @param <I>
+	 * 		Image pixel type
+	 *
+	 * @return True if the image labeling has intersecting labels, false otherwise.
+	 */
+	static < T, I extends IntegerType< I > > boolean hasIntersectingLabels( ImgLabeling< T, I > img )
```
imagejan, line 56: I'd vote for migrating this functionality upstream into `imglib2-roi`, e.g. in [`OverlappingLabels`](https://github.com/imglib/imglib2-roi/blob/f1e5daae75cab827790293b0bd9c9859f160d7c8/src/main/java/net/imglib2/roi/labeling/OverlappingLabels.java#L55) or some static utility class.

imagejan (commented):
Thanks @jdeschamps for this work! I think some of the basic functionality should actually live in `imglib2-roi` where we have already similar methods related to overlapping labels.

src/main/java/net/imglib2/algorithm/metrics/segmentation/SegmentationHelper.java
jdeschamps: See https://github.com/imglib/imglib2-roi/pull/59

src/main/java/net/imglib2/algorithm/metrics/segmentation/SegmentationHelper.java
jdeschamps: I've moved the methods to imglib2-roi, see this PR: https://github.com/imglib/imglib2-roi/pull/59

jdeschamps:
Right now this PR has to wait for the resolution of https://github.com/imglib/imglib2-roi/pull/59 !

jdeschamps:
New commits:
- Compatibility with the newly merged imglib/imglib2-roi#59 
- Add normalized _root mean square error (NRMSE)_

This PR is dependent on imglib2-roi **0.11.1**-SNAPSHOT (currently **0.10.4** in the scijava-pom 30.0.0 and **0.11.0** in the scijava-pom 32.0.0-SNAPSHOT)

ctrueden:
@jdeschamps Releasing imglib2-roi 0.12.0 now.

tpietzsch:
@jdeschamps I rebased on current master (3dcd24c89176d3806fd12d0334da442a968feb0c) and added a few easy simplifications. I will add more comments inline, but I think in principle this is fine as it is. Simplifications/optimizations can be made later.

The only real problem I see is `MunkresKuhnAlgorithm` (adapted from trackmate) which is under GPL license. This is incompatible with the BSD license of imglib2-algorithm. So either
- this class (and `MultiMetrics`, `LazyMultiMetrics` that use it) should be moved to [imglib2-algorithm-gpl](https://github.com/imglib/imglib2-algorithm-gpl), or
- we should obtain permission to relicense it under BSD, or
- this should be swapped with a different, compatibly licensed, implementation of the hungarian algorithm.
Whatever you prefer.

jdeschamps:
I get that!

I'd be in favor of relicensing. We should then ask @dscho whether he agrees to relicense MunkresKuhnAlgorithm.java from GPLv3 to BSD.

Do you know from the top of your head an alternative implementation? Otherwise I'd wait a bit and start looking for one in case @dscho disagrees.

dscho:
I am fine with the proposed relicensing!

tpietzsch:
Awesome! Thank you Johannes! @dscho

tpietzsch:
finally...

PR #94, opened 2022-02-25
Add Op discovery annotations to ImgLib2-algorithm

gselzer:
This PR uses [therapi-runtime-javadoc](https://github.com/dnault/therapi-runtime-javadoc) to declare all static functionality in imglib2-alrogithm as Ops. See [this test](https://github.com/scijava/incubator/blob/6d185b1585ceb27adb695b5a52c72ad38e25ad5c/imglib/imglib2-algorithm-optest/src/test/java/net/imglib2/algorithm/optest/OpTest.java#L17) for how these Ops can be called, and the module `imglib/imglib2-algorithm-optest` for an example of the structure required to call these annotated methods as Ops.

A new dependency on `therapi-runtime-javadoc-scribe` is added to the maven compiler plugin. *For now*, this was added to the imglib2-algorithm POM. Once `pom-scijava-base` 15.0.0 is propagated to imglib2-algorithm, we can remove almost everything added to this POM by this PR. 

The only line needed once we depend on that version of `pom-scijava-base` is [this line](https://github.com/imglib/imglib2-algorithm/blob/4f81a788017d76587c73326d2d52367489a59ddf/pom.xml#L203), which is responsible for informing therapi-runtime-javadoc-scribe of the set of packages it should process. This line must be changed once we bump the version, as the property used is under a different name.

Using `therapi-runtime-javadoc-scribe`, JSON files are generated for each file containing javadoc in each package specified by `therapi.packages`

Open questions:
* Op naming: I have not yet come up with a simple answer to this. I thought it would be nice to name the Ops using the method name, but some methods across different classes have the same name. We could name them using a camel-case `"class.method_name"`, which might match well with how these methods would be called from a static context...
* Which packages should be passed to therapi? We could just pass all packages, but this would retain javadoc that would not be used. We could also only pass packages with Op declarations, but it would be easy to forget to add new packages to this property when ops are declared in new packages. I tend to like the latter a little better, but don't feel too strongly.
* What to do with static methods not conforming to an Op? Many static methods are essentially [`org.scijava.function.Computer`](https://github.com/scijava/incubator/blob/main/scijava/scijava-function/src/main/java/org/scijava/function/Computers.java)s with their output parameter in a position that is **not** the last one; for these, I have added another method with a different parameter ordering. Other methods do not conform to Op types, such as the [`HessianMatrix`](https://github.com/imglib/imglib2-algorithm/blob/99ed0d777a2ede9b14501d32e1f8cb6f57368854/src/main/java/net/imglib2/algorithm/gradient/HessianMatrix.java#L91) methods. We could easily write a new `FunctionalInterface` for these guys, but they wouldn't have the support of `OpBuilder`, adaptation, etc.

TODO:
* [x] The build seems to fail in Java 8, but it runs just fine in Java 11. I still need to figure out why that is... :thinking:

### DIFF
--- a/src/main/java/net/imglib2/algorithm/componenttree/mser/MserTree.java
+++ b/src/main/java/net/imglib2/algorithm/componenttree/mser/MserTree.java
@@ -105,7 +105,8 @@ public final class MserTree< T extends Type< T > > implements ComponentForest< M
 	 * {@link #buildMserTree(RandomAccessibleInterval, RealType, long, long, double, double, ImgFactory, boolean)}
 	 * using an {@link ArrayImgFactory} or {@link CellImgFactory} depending on
 	 * input image size.
-	 * 
+	 *
+	 * @implNote op name='create.mserTree',type=Function
 	 * @param input
 	 *            the input image.
 	 * @param delta
@@ -133,7 +134,8 @@ public final class MserTree< T extends Type< T > > implements ComponentForest< M
 	 * {@link #buildMserTree(RandomAccessibleInterval, RealType, long, long, double, double, ImgFactory, boolean)}
 	 * using an {@link ArrayImgFactory} or {@link CellImgFactory} depending on
 	 * input image size.
-	 * 
+	 *
+	 * @implNote op name='create.mserTree',type=Function
 	 * @param input
 	 *            the input image.
 	 * @param delta
@@ -159,7 +161,8 @@ public final class MserTree< T extends Type< T > > implements ComponentForest< M
 
 	/**
 	 * Build a MSER tree from an input image.
-	 * 
+	 *
+	 * @implNote op name='create.mserTree',type=Function
 	 * @param input
 	 *            the input image.
 	 * @param delta
@@ -198,7 +201,8 @@ public final class MserTree< T extends Type< T > > implements ComponentForest< M
 	 * {@link #buildMserTree(RandomAccessibleInterval, ComputeDelta, long, long, double, double, ImgFactory, Type, Comparator)}
 	 * using an {@link ArrayImgFactory} or {@link CellImgFactory} depending on
 	 * input image size.
-	 * 
+	 *
+	 * @implNote op name='create.mserTree',type=Function
 	 * @param input
 	 *            the input image.
 	 * @param computeDelta
@@ -226,7 +230,8 @@ public final class MserTree< T extends Type< T > > implements ComponentForest< M
 
 	/**
 	 * Build a MSER tree from an input image.
-	 * 
+	 *
+	 * @implNote op name='create.mserTree',type=Function
 	 * @param input
 	 *            the input image.
 	 * @param computeDelta
--- a/src/main/java/net/imglib2/algorithm/componenttree/pixellist/PixelListComponentTree.java
+++ b/src/main/java/net/imglib2/algorithm/componenttree/pixellist/PixelListComponentTree.java
@@ -74,6 +74,7 @@ public final class PixelListComponentTree< T extends Type< T > > implements Comp
 	 * using an {@link ArrayImgFactory} or {@link CellImgFactory} depending on
 	 * input image size.
 	 *
+	 * @implNote op name='create.pixelListComponentTree',type=Function
 	 * @param input
 	 *            the input image.
 	 * @param type
@@ -92,6 +93,7 @@ public final class PixelListComponentTree< T extends Type< T > > implements Comp
 	/**
 	 * Build a component tree from an input image.
 	 *
+	 * @implNote op name='create.pixelListComponentTree',type=Function
 	 * @param input
 	 *            the input image.
 	 * @param type
@@ -120,6 +122,7 @@ public final class PixelListComponentTree< T extends Type< T > > implements Comp
 	 * using an {@link ArrayImgFactory} or {@link CellImgFactory} depending on
 	 * input image size.
 	 *
+	 * @implNote op name='create.pixelListComponentTree',type=Function
 	 * @param input
 	 *            the input image.
 	 * @param maxValue
@@ -138,6 +141,7 @@ public final class PixelListComponentTree< T extends Type< T > > implements Comp
 	/**
 	 * Build a component tree from an input image.
 	 *
+	 * @implNote op name='create.pixelListComponentTree',type=Function
 	 * @param input
 	 *            the input image.
 	 * @param maxValue
--- a/src/main/java/net/imglib2/algorithm/convolution/fast_gauss/FastGauss.java
+++ b/src/main/java/net/imglib2/algorithm/convolution/fast_gauss/FastGauss.java
@@ -88,11 +88,23 @@ public class FastGauss
 		return new LineConvolution<>( new FastGaussConvolverRealType( sigma ), direction );
 	}
 
+	/**
+	 * @param sigmas the standard deviations of the Gaussian blur (in each dimension)
+	 * @param input the input data to blur
+	 * @param output the preallocated output buffer
+	 * @implNote op name='filter.gauss', type=Computer
+	 */
 	public static void convolve( final double[] sigmas, final RandomAccessible< ? extends RealType< ? > > input, final RandomAccessibleInterval< ? extends RealType< ? > > output )
 	{
 		convolution( sigmas ).process( input, output );
 	}
 
+	/**
+	 * @param sigma the standard deviation of the Gaussian blur (in all dimensions)
+	 * @param input the input data to blur
+	 * @param output the preallocated output buffer
+	 * @implNote op name='filter.gauss', type=Computer
+	 */
 	public static void convolve( final double sigma, final RandomAccessible< ? extends RealType< ? > > input, final RandomAccessibleInterval< ? extends RealType< ? > > output )
 	{
 		convolution( sigma ).process( input, output );
--- a/src/main/java/net/imglib2/algorithm/convolution/kernel/Kernel1D.java
+++ b/src/main/java/net/imglib2/algorithm/convolution/kernel/Kernel1D.java
@@ -56,6 +56,8 @@ public class Kernel1D
 	 * @param halfKernel
 	 *            the upper half (starting at the center pixel) of the symmetric
 	 *            convolution kernel.
+	 * @return a {@link Kernel1D} used for one dimensional convolutions
+	 * @implNote op name='create.kernel1DSymmetric'
 	 */
 	public static Kernel1D symmetric( final double... halfKernel )
 	{
@@ -68,6 +70,9 @@ public class Kernel1D
 	/**
 	 * Similar to {@link #symmetric(double[])} but creates an array of
 	 * one-dimensional convolution kernels.
+	 * @param halfKernels the upper halves (starting at the center pixel) of the symmetric convolution kernels
+	 * @return an array of {@link Kernel1D}s used for one dimensional convolutions
+	 * @implNote op name='create.kernel1DSymmetric'
 	 */
 	public static Kernel1D[] symmetric( final double[][] halfKernels )
 	{
@@ -82,6 +87,8 @@ public class Kernel1D
 	 * @param originIndex
 	 *            the index of the array element which is the origin of the
 	 *            kernel
+	 * @return an asymmetric {@link Kernel1D} used for one dimensional convolutions
+	 * @implNote op name='create.kernel1DAsymmetric'
 	 */
 	public static Kernel1D asymmetric( final double[] fullKernel, final int originIndex )
 	{
@@ -94,6 +101,9 @@ public class Kernel1D
 	/**
 	 * Creates a one-dimensional asymmetric convolution kernel, where the origin
 	 * of the kernel is in the middle.
+	 * @param kernel the raw data of the symmetric convolution kernel
+	 * @return a {@link Kernel1D} used for one dimensional convolutions
+	 * @implNote op name='create.kernel1DCentralAsymmetric'
 	 */
 	public static Kernel1D centralAsymmetric( final double... kernel )
 	{
@@ -103,6 +113,13 @@ public class Kernel1D
 	/**
 	 * Similar to {@link #asymmetric(double[], int)} but creates an array of
 	 * one-dimensional convolution kernels.
+	 * @param fullKernels
+	 *            arrays containing the values of each kernel
+	 * @param originIndices
+	 *            the indices of the array elements at the origin of each
+	 *            kernel
+	 * @return an array of {@link Kernel1D}s used for one dimensional convolutions
+	 * @implNote op name='create.kernel1DAsymmetric'
 	 */
 	public static Kernel1D[] asymmetric( final double[][] fullKernels, final int[] originIndices )
 	{
@@ -113,6 +130,9 @@ public class Kernel1D
 	/**
 	 * Similar to {@link #centralAsymmetric(double...)} but creates an array of
 	 * one-dimensional convolution kernels.
+	 * @param kernels the upper halves (starting at the center pixel) of the symmetric convolution kernels
+	 * @return an array of {@link Kernel1D}s used for one dimensional convolutions
+	 * @implNote op name='create.kernel1DCentralAsymmetric'
 	 */
 	public static Kernel1D[] centralAsymmetric( final double[][] kernels )
 	{
--- a/src/main/java/net/imglib2/algorithm/dog/DifferenceOfGaussian.java
+++ b/src/main/java/net/imglib2/algorithm/dog/DifferenceOfGaussian.java
@@ -69,6 +69,44 @@ public class DifferenceOfGaussian
 	 * {@link #DoG(double[], double[], RandomAccessible, RandomAccessible, RandomAccessibleInterval, ExecutorService)}
 	 * .
 	 * </p>
+	 * This method differs from
+	 * {@link #DoG(double[], double[], RandomAccessible, RandomAccessibleInterval, ExecutorService)}
+	 * only in that its parameter order is tailored to an Op. The output comes
+	 * last, and the primary input (the input image) comes first.
+	 *
+	 * @implNote op name="filter.DoG", type=Computer
+	 * @param input
+	 *            the input image extended to infinity (or at least covering the
+	 *            same interval as the dog result image, plus borders for
+	 *            convolution).
+	 * @param sigmaSmaller
+	 *            stddev (in every dimension) of smaller Gaussian.
+	 * @param sigmaLarger
+	 *            stddev (in every dimension) of larger Gaussian.
+	 * @param service
+	 *            service providing threads for multi-threading
+	 * @param dog
+	 *            the Difference-of-Gaussian result image.
+	 */
+	public static < I extends NumericType< I >, T extends NumericType< T > & NativeType< T > > void DoG(
+			final RandomAccessible< I > input,
+			final double[] sigmaSmaller,
+			final double[] sigmaLarger,
+			final ExecutorService service,
+			final RandomAccessibleInterval< T > dog)
+	{
+		DoG( sigmaSmaller, sigmaLarger, input, dog, service );
+	}
+
+	/**
+	 * Compute the difference of Gaussian for the input. Input convolved with
+	 * Gaussian of sigmaSmaller is subtracted from input convolved with Gaussian
+	 * of sigmaLarger (where {@code sigmaLarger > sigmaSmaller}).
+	 * <p>
+	 * Creates an appropriate temporary image and calls
+	 * {@link #DoG(double[], double[], RandomAccessible, RandomAccessible, RandomAccessibleInterval, ExecutorService)}
+	 * .
+	 * </p>
 	 *
 	 * @param sigmaSmaller
 	 *            stddev (in every dimension) of smaller Gaussian.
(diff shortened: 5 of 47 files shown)


gselzer:
A running list of `static` methods that do not conform to functional interfaces in scijava-function:
* The [`HessianMatrix`](https://github.com/imglib/imglib2-algorithm/blob/99ed0d777a2ede9b14501d32e1f8cb6f57368854/src/main/java/net/imglib2/algorithm/gradient/HessianMatrix.java#L91) methods
* The [`ConnectedComponents.labelAllConnectedComponents`](https://github.com/imglib/imglib2-algorithm/blob/99ed0d777a2ede9b14501d32e1f8cb6f57368854/src/main/java/net/imglib2/algorithm/labeling/ConnectedComponents.java#L176) that return an `int`

gselzer:
@ctrueden @tpietzsch I've fully updated this PR such that it now uses (the unreleased) SciJava Ops Indexer. I can now automatically discover all annotated static methods in an environment containing both a local build of imglib2-algorithm and a local build of the SciJava Incubator. The PR will not build until we release the SciJava Ops Indexer, so I've left it as a draft, but bringing this work up-to-date made me excited for the possibilities of SciJava Ops!

hinerm:
@tpietzsch In [this PR](https://github.com/scijava/scijava/pull/236) we are discussing the minimum Java version to index libraries with Ops. I'm wondering if there is a need to keep `imglib2-algorithm` at Java 8, or if going to Java 11 is an option?

tpietzsch:
@hinerm I don't think there is anything that would break with higher Java versions. As soon as Fiji finally moves to Java 21, (all of) ImgLib2 will too.

gselzer:
TODO:

There are some methods whose signatures do not conform to the Ops style (e.g. [here](https://github.com/imglib/imglib2-algorithm/pull/94/files#diff-cd262144ec53f9c735848b62c4db8d349c5e226b6d15bc43bdc2d703bfe4b2d1R124)), where I've written a wrapper method that has been opified. The question, then, is what to do with the original methods. Should we consider deprecating them, @ctrueden?

src/main/java/net/imglib2/algorithm/componenttree/mser/MserTree.java
```diff
@@ -105,7 +105,8 @@ public final class MserTree< T extends Type< T > > implements ComponentForest< M
 	 * {@link #buildMserTree(RandomAccessibleInterval, RealType, long, long, double, double, ImgFactory, boolean)}
 	 * using an {@link ArrayImgFactory} or {@link CellImgFactory} depending on
 	 * input image size.
-	 *
```
gselzer, line 108: Opinions on whether these should be Ops? I'm not totally sure what they're used for...

src/main/java/net/imglib2/algorithm/componenttree/pixellist/PixelListComponentTree.java
```diff
@@ -74,6 +74,7 @@ public final class PixelListComponentTree< T extends Type< T > > implements Comp
 	 * using an {@link ArrayImgFactory} or {@link CellImgFactory} depending on
 	 * input image size.
 	 *
+	 * @implNote op name='create.pixelListComponentTree',type=Function
 	 * @param input
```
gselzer, line 78: Like above, should these be ops?

src/main/java/net/imglib2/algorithm/dog/DifferenceOfGaussian.java
```diff
@@ -97,6 +135,44 @@ public static < I extends NumericType< I >, T extends NumericType< T > & NativeT
...
+	 *            stddev (in every dimension) of larger Gaussian.
+	 * @param tmp
+	 *            temporary image, must at least cover the same interval as the
+	 *            dog result image.
+	 * @param service
+	 *            how many threads to use for the computation.
+	 * @param dog
+	 *            the Difference-of-Gaussian result image.
+	 */
+	public static < I extends NumericType< I >, T extends NumericType< T > & NativeType< T > > void DoG(
```
gselzer, line 165: I wrote these wrappers around the original methods to make the signatures more op-like. What should we do with the older ones? Nothing? Deprecate them?

src/main/java/net/imglib2/algorithm/edge/SubpixelEdgelDetection.java
```diff
@@ -73,7 +73,8 @@ public class SubpixelEdgelDetection
 	 * <p>
 	 * Note: The input image type must be a signed type! Otherwise gradient
 	 * computation will not work.
-	 * 
+	 *
+	 * @implNote op name='image.subpixelEdgels', type=Function
```
gselzer, line 77: This maybe belongs better in the `features` namespace?

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -136,6 +138,7 @@ public static long[] getVotespaceSize( final Dimensions dimensions, final int nT
 	 * Returns the size of the voteSpace output image given desired {@code nRho}
 	 * and {@code nTheta} values.
 	 *
+	 * @implNote op names='features.hough.getVotespaceSize', type=Function
```
gselzer, line 141: The `features.hough.getVotespaceSize` is not idea. At a minimum, change it to `features.hough.linearVotespaceSize`, but I feel like there is a better classification here. Thoughts, anyone?

src/main/java/net/imglib2/algorithm/localextrema/LocalExtrema.java
```diff
@@ -167,6 +167,7 @@ public static < P, T > ArrayList< P > findLocalExtrema( final RandomAccessibleIn
 	 * {@code source} accordingly. The returned coordinate list is valid
 	 * for the original {@code source}.
 	 *
+	 * @implNote op name='image.localExtrema', type=Function
```
gselzer, line 170: Should this belong in the `features` namespace?

src/main/java/net/imglib2/algorithm/morphology/StructuringElements.java
```diff
@@ -88,6 +88,7 @@ public class StructuringElements
 	 * This methods relies on heuristics to determine automatically what
 	 * decomposition level to use.
 	 *
+	 * @implNote op name='create.disk', type=Function
```
gselzer, line 91: Should these be Ops?

src/main/java/net/imglib2/algorithm/util/Grids.java
```diff
@@ -213,6 +213,7 @@ public static void forEachOffset(
 	 * Get all blocks of size {@code blockSize} contained within an interval
 	 * specified by {@code dimensions} and their positions within a cell grid.
 	 *
+	 * @implNote op name='grid.partition', type=Function
```
gselzer, line 216: Should these be Ops? To which namespace should they belong?

src/main/java/net/imglib2/algorithm/util/ParallelizeOverBlocks.java
```diff
@@ -61,6 +61,7 @@ public class ParallelizeOverBlocks
 	 *
 	 * Submit blocked tasks and wait for execution.
 	 *
+	 * @implNote op name='grid.parallelize', type=Function
```
gselzer, line 64: Again, should these be Ops? Is this a good namespace?

src/main/java/net/imglib2/algorithm/convolution/kernel/Kernel1D.java
```diff
@@ -56,6 +56,8 @@ public class Kernel1D
 	 * @param halfKernel
 	 *            the upper half (starting at the center pixel) of the symmetric
 	 *            convolution kernel.
+	 * @return a {@link Kernel1D} used for one dimensional convolutions
+	 * @implNote op name='create.kernel1DSymmetric'
```
gselzer, line 60: Should these be Ops? What is the right name?

gselzer (commented):
@ctrueden @elevans I'd appreciate your opinions here!

src/main/java/net/imglib2/algorithm/dog/DifferenceOfGaussian.java
ctrueden: Let's do nothing with the old ones. At most: put a `//` comment near them explaining why they aren't declared as Ops, if you think there is a chance someone will wonder about it later.

src/main/java/net/imglib2/algorithm/componenttree/mser/MserTree.java
ctrueden: https://github.com/imglib/imglib2-tests/blob/180e0f988c572518e0b35bedc558186a5b79be8a/src/test/java/net/imglib2/algorithm/componenttree/mser/MserTreeExample.java

ctrueden: Concepts of level sets, mathematical morphology, and connected component analysis are very related. We have CCA currently in `labeling`, but other morphology operators in `morphology`. What about [level sets](https://imagej.net/plugins/level-sets)? FWIW, I checked scikit-image for comparison, and use of level sets mostly takes place in `skimage.segmentation.morphsnakes`.

src/main/java/net/imglib2/algorithm/componenttree/pixellist/PixelListComponentTree.java
ctrueden: Can we use a sub-namespace of `create` here? `create.componentTree.pixelList`, `create.componentTree.mser`, and so on? It would match the package naming. In general, looking at the Java package naming is a good starting point for naming suggestions.

ctrueden: Actually... are these really "create" Ops? I think of `create` as merely allocating and returning an object instance, not doing any computation around an image. So this `PixelListComponentTree` shouldn't be a `create` Op at all, since it accepts an existing image as input. `morphology.componentTree.pixelList` maybe? I don't know this space of image algorithms well enough to feel qualified to suggest anything. We could use @tpietzsch's input here, since he wrote these classes. And/or just read the papers about these things.

src/main/java/net/imglib2/algorithm/convolution/fast_gauss/FastGauss.java
```diff
@@ -88,11 +88,23 @@ public class FastGauss
 		return new LineConvolution<>( new FastGaussConvolverRealType( sigma ), direction );
 	}
 
+	/**
+	 * @param sigmas the standard deviations of the Gaussian blur (in each dimension)
+	 * @param input the input data to blur
+	 * @param output the preallocated output buffer
+	 * @implNote op name='filter.gauss', type=Computer
```
ctrueden, line 95: Hmm, interesting. We may want to make a post on Image.sc asking for people's suggestions on what Ops should do for naming w.r.t. approximate vs. exact algorithms like this. Is it OK for `filter.gauss` to be approximate by default? Should the exact and approximate versions have different names? Or can we use a hint somehow?

src/main/java/net/imglib2/algorithm/convolution/kernel/Kernel1D.java
ctrueden: Would it make sense to group all the kernel creation ops into `create.kernel.` sub-namespace? E.g. `create.kernel.symmetric1D`? Or maybe `create.kernel1D.symmetric` since these all return `Kernel1D` objects. Not sure...

src/main/java/net/imglib2/algorithm/dog/DifferenceOfGaussian.java
```diff
@@ -97,6 +135,45 @@ public static < I extends NumericType< I >, T extends NumericType< T > & NativeT
...
+	 * Gaussian of sigmaSmaller is subtracted from input convolved with Gaussian
+	 * of sigmaLarger (where sigmaLarger &gt; sigmaSmaller).
+	 * <p>
+	 * This method differs from
+	 * {@link #DoG(double[], double[], RandomAccessible, RandomAccessible, RandomAccessibleInterval, ExecutorService)}
+	 * only in that its parameter order is tailored to an Op. The output comes
+	 * last, and the primary input (the input image) comes first.
+	 * </p>
+	 *
+	 * @implNote op name="filter.DoG", type=Computer
```
ctrueden, line 149: In scijava-ops-image, it's always `filter.dog` lowercase.

src/main/java/net/imglib2/algorithm/edge/SubpixelEdgelDetection.java
ctrueden: Sure, rename it.

src/main/java/net/imglib2/algorithm/fill/FloodFill.java
```diff
@@ -90,6 +90,7 @@ public class FloodFill
 	 *            input pixel type
 	 * @param <U>
 	 *            fill label type
+	 * @implNote op name='image.floodFill', type=Computer2
```
ctrueden, line 93: Let's move to `morphology`. That's where skimage puts flood fill.

src/main/java/net/imglib2/algorithm/gauss3/Gauss3.java
```diff
@@ -116,6 +117,7 @@ public static < S extends NumericType< S >, T extends NumericType< T > > void ga
 	 * {@link Parallelization} context. (By default, it will use the
 	 * {@link ForkJoinPool#commonPool() common ForkJoinPool})
 	 *
+	 * @implNote op name='filter.gauss', type=Computer
```
ctrueden, line 120: This is the exact Gaussian... need to decide how to name compared to the approximate one.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
ctrueden: Look to the method name for guidance. It's `...hough.HoughTransforms.getVotespaceSize`. The `get` is just a Javaism, and we need the `features.` Op prefix for organization, so that leaves `features.hough.votespaceSize`.

In other words: if this is only the *linear* votespace size calculation, and there are also *non-linear* ones, it is not only the name of the Op that will need to change, but also the static method name itself here, no?

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
```diff
@@ -150,6 +153,7 @@ public static long[] getVotespaceSize( final int nRho, final int nTheta )
 	/**
 	 * Pick vote space peaks with a {@link LocalExtrema}.
 	 *
+	 * @implNote op names='features.hough.pickPeaks', type=Function
```
ctrueden, line 156: Conversely from the `votespaceSize` naming question: the method here is named `pickLinePeaks`. Is it possible for there to be *non-line* peaks in a different context of Hough? If so, we should leave the Op named `features.hough.pickLinePeaks` for consistency. Otherwise, we could consider renaming the Java method.

src/main/java/net/imglib2/algorithm/morphology/StructuringElements.java
ctrueden: Also: should this `create` Op also be an `engine.create` alias? If not: does it really belong in the `create` namespace?"

src/main/java/net/imglib2/algorithm/morphology/StructuringElements.java
```diff
@@ -145,6 +146,7 @@ else if ( radius < 17 )
 	 * select the level of approximation. For other dimensionalities, no
 	 * optimization are available yet and the parameter is ignored.
 	 *
+	 * @implNote op name='create.disk', type=Function
```
ctrueden, line 149: `engine.create` alias?

src/main/java/net/imglib2/algorithm/morphology/StructuringElements.java
```diff
@@ -291,6 +293,7 @@ else if ( decomposition == 8 || decomposition == 4 || decomposition == 6 )
 	 * dimensionality and target dimensionality do not match. Non-decomposed
 	 * version are dimension-generic.
 	 *
+	 * @implNote op name='create.square', type=Function
```
ctrueden, line 296: `engine.create` alias?

src/main/java/net/imglib2/algorithm/morphology/StructuringElements.java
```diff
@@ -340,6 +343,7 @@ public static final List< Shape > square( final int radius, final int dimensiona
 	 * This method determines whether it is worth returning a decomposed strel
 	 * based on simple heuristics.
 	 *
+	 * @implNote op name='create.square', type=Function
```
ctrueden, line 346: `engine.create` alias?

src/main/java/net/imglib2/algorithm/morphology/TopHat.java
```diff
@@ -78,7 +78,8 @@ public class TopHat
 	 * decomposition. Each shape is processed in order as given in the list. If
 	 * the list is empty, the source image is returned.
 	 * <p>
-	 * 
+	 *
+	 * @implNote op names='morphology.topHat, morphology.whiteTopHat', type=Function
```
ctrueden, line 82: Let's drop the `morphology.topHat` alias, in favor of `whiteTopHat` always, for symmetry with `blackTopHat`. That's what scikit-image does.

src/main/java/net/imglib2/algorithm/util/Grids.java
ctrueden: Let's hold off on the grid functions for now.

ctrueden (changes requested):
OK, I expressed some opinions. Hope it helps.

src/main/java/net/imglib2/algorithm/dog/DifferenceOfGaussian.java
gselzer: I already wrote a comment on the ones that *do* have Ops - that's probably enough?

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
gselzer: For now, we have decided not to opify hough transforms.

src/main/java/net/imglib2/algorithm/morphology/TopHat.java
gselzer: Heh, SciJava Ops Image currently *only* uses `morphology.topHat`. I agree that symmetry would be good, but I'm also tempted to leave in the aliases. Maybe we can deprecate that soon?

src/main/java/net/imglib2/algorithm/morphology/StructuringElements.java
gselzer: I'm not a huge fan of this. Most of these shapes are created by, and return, the same types, yielding a lot of conflicting results.

Honestly, let's maybe just de-opify these methods for now, as they are not used...

gselzer: On second thought, I actually think that these Ops would be pretty helpful, **except** for the fact that they return a `List<Shape>`, **not** the `Shape` itself. We should probably expose the mechanisms to just create a plain `Shape`...

src/main/java/net/imglib2/algorithm/morphology/StructuringElements.java
gselzer: While I like the idea of having `create` Ops for `Shapes`, I again don't think that these methods are quite general enough, so I'm just not going to opify them, *yet*.

src/main/java/net/imglib2/algorithm/fill/FloodFill.java
gselzer: Good idea 👍

src/main/java/net/imglib2/algorithm/convolution/fast_gauss/FastGauss.java
gselzer: * As we talked about Op naming, the fundamental idea is that numerical equivalence is not required for two Ops to share a name. So I think that both should be named `filter.gauss`. I don't like any differentiation between the names because then you're being explicit about implementation details, contrary to the Ops design.
* Without additional changes, `filter.gauss` won't be approximate by default but would be approximate *always*, which is probably bad. Actually, right now `filter.gauss(sigma, source, target)` won't work because there's a conflict, and one *will* have to be prioritized. I'd lean towards prioritizing the approximate one for speed, and we should likely provide a hint that could be used to avoid this Op. There are likely additional scenarios where this would be useful, such as convolution in SciJava Ops Image.

As is, we need to prioritize one of the Ops before we merge this PR, and we may want to implement the Precision Hint upstream beforehand as well.

src/main/java/net/imglib2/algorithm/componenttree/pixellist/PixelListComponentTree.java
gselzer: It actually seems to be like the tree best belongs to the `features` namespace, as it is my understanding that component trees themselves have distilled down a 2D function into a tree, where the position is lost. This seems more like feature extraction than morphology.

src/main/java/net/imglib2/algorithm/hough/HoughTransforms.java
ctrueden: We did? Maybe I misunderstood the scope of these methods, then...

src/main/java/net/imglib2/algorithm/convolution/fast_gauss/FastGauss.java
ctrueden: I like the idra of a precision hint. It is more general than only this Op. E.g. arbitrary precision arithmetic.

PR #107, opened 2025-02-26
Distance transform propagate labels

bogovicj:
Working toward propagating a set of labels while performing a distance transform.

`transformPropagateLabels(distance, labels, distanceType)`

behaves identically to `transform` with respect to the `distance` argument, but also
modifies the input labels such that the label at any given point is given by the label
with the closes distance to that point (where the distances are given by 

For example, for 1D inputs `distance` and `labels` given by:
```
distance = [ 9.0, 9.0, 1.0, 9.0, 9.0, 9.0,  1.5, 9.0 ]  
labels   = [   4,   2,   1,   0,   8,   0,    2,   0 ]
```
will modify labels to be:
```
labels   = [   1,   1,   1,   1,   1,   2,    2,   2 ]
```

### DIFF
--- a/src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
+++ b/src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
@@ -738,7 +738,7 @@ public class DistanceTransform
 			final DISTANCE_TYPE distanceType,
 			final double... weights )
 	{
-		final U maxVal = tmp.getType().copy();
+		final U maxVal = tmp.getType().createVariable();
 		maxVal.setReal( maxVal.getMaxValue() );
 		final Converter< B, U > converter = new BinaryMaskToCost<>( maxVal );
 		final RandomAccessible< U > converted = Converters.convert( source, converter, maxVal.createVariable() );
@@ -792,7 +792,7 @@ public class DistanceTransform
 			final int nTasks,
 			final double... weights ) throws InterruptedException, ExecutionException
 	{
-		final U maxVal = tmp.getType().copy();
+		final U maxVal = tmp.getType().createVariable();
 		maxVal.setReal( maxVal.getMaxValue() );
 		final Converter< B, U > converter = new BinaryMaskToCost<>( maxVal );
 		final RandomAccessible< U > converted = Converters.convert( source, converter, maxVal.createVariable() );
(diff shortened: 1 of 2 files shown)


src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
```diff
@@ -963,7 +966,7 @@ public static < B extends BooleanType< B >, U extends RealType< U >, V extends R
 			final ExecutorService es,
 			final int nTasks ) throws InterruptedException, ExecutionException
 	{
-		final U maxVal = Util.getTypeFromInterval( tmp ).createVariable();
+		final U maxVal = tmp.getType().copy();
```
axtimwalde: Can this be getType().createVariable() ?  We do not need the value.

src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
```diff
@@ -916,7 +919,7 @@ public static < B extends BooleanType< B >, U extends RealType< U >, V extends R
 			final RandomAccessibleInterval< V > target,
 			final Distance d )
 	{
-		final U maxVal = Util.getTypeFromInterval( tmp ).createVariable();
+		final U maxVal = tmp.getType().copy();
```
axtimwalde: Can this be getType().createVariable() ?  We do not need the value.

src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
```diff
@@ -789,7 +792,7 @@ public static < B extends BooleanType< B >, U extends RealType< U >, V extends R
 			final int nTasks,
 			final double... weights ) throws InterruptedException, ExecutionException
 	{
-		final U maxVal = Util.getTypeFromInterval( tmp ).createVariable();
+		final U maxVal = tmp.getType().copy();
```
axtimwalde: Can this be getType().createVariable() ?  We do not need the value.

src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
```diff
@@ -734,8 +738,7 @@ public static < B extends BooleanType< B >, U extends RealType< U >, V extends R
 			final DISTANCE_TYPE distanceType,
 			final double... weights )
 	{
-
-		final U maxVal = Util.getTypeFromInterval( tmp ).createVariable();
+		final U maxVal = tmp.getType().copy();
```
axtimwalde: Can this be getType().createVariable() ?  We do not need the value.

src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
axtimwalde: Done.

axtimwalde (approved):
Done

Issue #108, opened 2025-03-06
DistanceTransform throws Exception with certain weights


bogovicj:
This code

```java
double M = Double.MAX_VALUE;
double[] data = new double[] {
		0, M, M, M,
		M, 0, 0, 0
};

double wx = 0.1;
double wy = 1.0
ArrayImg<DoubleType, DoubleArray> dists = ArrayImgs.doubles(data, 4, 2);
DistanceTransform.transform(dists, new EuclidianDistanceAnisotropic(wx, wy));
```

throws:

```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: -1
	at net.imglib2.img.basictypeaccess.array.AbstractLongArray.getValue(AbstractLongArray.java:61)
	at net.imglib2.type.numeric.integer.GenericLongType.getLong(GenericLongType.java:135)
	at net.imglib2.type.numeric.integer.LongType.get(LongType.java:93)
	at net.imglib2.algorithm.morphology.distance.DistanceTransform.transformSingleColumn(DistanceTransform.java:1193)
	at net.imglib2.algorithm.morphology.distance.DistanceTransform.transformAlongDimension(DistanceTransform.java:1128)
	at net.imglib2.algorithm.morphology.distance.DistanceTransform.transform(DistanceTransform.java:541)
	at net.imglib2.algorithm.morphology.distance.DistanceTransform.transform(DistanceTransform.java:455)
	at net.imglib2.algorithm.morphology.distance.DistanceTransform.transform(DistanceTransform.java:393)
```

weights less than 1.0 seem to be the cause of this.

PR #109, opened 2025-03-06
fix/test: DistanceTransform array out of bounds exception

bogovicj:
* adds tests that reproduce the error in #108

### DIFF
--- a/src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
+++ b/src/main/java/net/imglib2/algorithm/morphology/distance/DistanceTransform.java
@@ -1187,7 +1187,7 @@ public class DistanceTransform
 			final double sourceAtPosition = source.get( position ).getRealDouble();
 			double s = d.intersect( envelopeIndexAtK, source.get( envelopeIndexAtK ).getRealDouble(), position, sourceAtPosition, dim );
 
-			for ( double envelopeValueAtK = envelopeIntersectLocation.get( k ).get(); s <= envelopeValueAtK; envelopeValueAtK = envelopeIntersectLocation.get( k ).get() )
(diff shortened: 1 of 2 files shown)


bogovicj:
The second commit to this PR substantially speeds things up (40% - 100%), but using primitive arrays for parts of the  lowest level methods `transformAlongDimension` and `transformAlongDimensionPropagateLabels`

Because this implementation makes an attempt to support very large images, we fall back to the old implementation if any dimension is longer than `Integer.MAX_VALUE`. 

That commit also includes a benchmark, some of the results of which are shown below:


# Selected benchmark results
 
## 2D

### old single thread
```
Benchmark                                            Mode  Cnt    Score   Error  Units
DistanceTransformBenchmark.voronoiDistanceTransform  avgt   25  484.629 ± 6.081  ms/op
```
### new single thread
```
Benchmark                                            Mode  Cnt    Score   Error  Units
DistanceTransformBenchmark.voronoiDistanceTransform  avgt   25  295.930 ± 3.423  ms/op
```

### old multi-thread (8)
```
Benchmark                                            Mode  Cnt    Score   Error  Units
DistanceTransformBenchmark.voronoiDistanceTransform  avgt   25  170.793 ± 1.800  ms/op
```
### new multi-thread (8)
```
Benchmark                                            Mode  Cnt    Score   Error  Units
DistanceTransformBenchmark.voronoiDistanceTransform  avgt   25  118.340 ± 2.338  ms/op
```

## 3D

### old single threaded

```
Benchmark                                            Mode  Cnt     Score    Error  Units
DistanceTransformBenchmark.distanceTransform         avgt   25  1167.336 ± 14.451  ms/op
DistanceTransformBenchmark.voronoiDistanceTransform  avgt   25  1852.702 ± 20.670  ms/op
```

### new single threaded
```
Benchmark                                            Mode  Cnt     Score    Error  Units
DistanceTransformBenchmark.distanceTransform         avgt   25   605.162 ±  6.575  ms/op
DistanceTransformBenchmark.voronoiDistanceTransform  avgt   25  1315.934 ± 22.510  ms/op
```

